PR #17, opened 2023-02-13
Feature: Multi-select prompt

EduSantosBrito:
Hey!

I loved the package :) I think it needs a multi-select prompt, it will be great!

I followed the pattern that you create (any thoughts are welcome).

Maybe a good improvement is returning `T | T[]` instead of `string` from render function on close.


natemoo-re (approved):
Thank you so much for the PR, this is awesome!

PR #18, opened 2023-02-13
fix: `tsconfig.json` and de-dup `devDeps` for project

cpreston321:
# Fix

In this PR I resolve deduping `tsconfig.json` and `devDeps`. Since TS is throughout the project and a monorepo it's better to have the base config live at the root and when you `@types/..` you can then add them to the root project and it will be incorporated throughout the project.

if you have any questions, let me know! 

_P.S_ Everything works the same just makes packages feel lighter.

Thanks,
Christian


natemoo-re (approved):
Awesome, I had been meaning to do this. Thanks!

PR #19, opened 2023-02-13
fix: move `demo` to `examples/`

cpreston321:
# Fix

In this PR since demo is more of example it made more since for it to be in the `examples/` folder. I included it within the pnpm-workspace, so now you can add more examples if you needed too.


natemoo-re (approved):
🙌 Thanks so much!

PR #23, opened 2023-02-14
feat: add `eslint` formatter.

cpreston321:
# ESLint

What does this do ?

- Format code using ESLint just by clicking CMD+S
- See errors within TS a head of time.
- @antfu/eslint-config preset [Repo Here](https://github.com/antfu/eslint-config)
- Make sure everyone has the same tabs, spaces, semi, etc.. to keep formatting consistent.

## Additionally

- Added `lint`, `lint:fix` and `cleanup` to `package.json`
  - `cleanup` - Easily removes all `node_modules/**` in each directory with one command.

## Config VS Code auto fix

Install [VS Code ESLint extension](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) and create `.vscode/settings.json`

```json
{
  "prettier.enable": false,
  "editor.formatOnSave": false,
  "editor.codeActionsOnSave": {
    "source.fixAll.eslint": true
  }
}
```

This is also opinionated, so if you don't want to add this I understand.

Thanks,
CP 🚀

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,3 @@
+{
+  "extends": "@antfu"
+}
--- a/build.preset.ts
+++ b/build.preset.ts
@@ -10,6 +10,6 @@ export default definePreset({
     esbuild: {
       minify: true,
       sourceMap: true,
-    }
-  }
+    },
+  },
 })
--- a/examples/basic/index.ts
+++ b/examples/basic/index.ts
@@ -1,87 +1,88 @@
+/* eslint-disable no-console */
+import { setTimeout } from 'node:timers/promises'
 import {
-  text,
-  select,
+  cancel,
   confirm,
   intro,
-  outro,
-  cancel,
-  spinner,
   isCancel,
   multiselect,
   note,
-} from "@clack/prompts";
-import color from "picocolors";
-import { setTimeout } from "node:timers/promises";
+  outro,
+  select,
+  spinner,
+  text,
+} from '@clack/prompts'
+import color from 'picocolors'
 
 async function main() {
-  console.clear();
+  console.clear()
 
-  await setTimeout(1000);
+  await setTimeout(1000)
 
-  intro(`${color.bgCyan(color.black(" create-app "))}`);
+  intro(`${color.bgCyan(color.black(' create-app '))}`)
 
   const dir = await text({
-    message: "Where should we create your project?",
-    placeholder: "./sparkling-slop",
-  });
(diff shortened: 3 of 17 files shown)


natemoo-re:
Hey sorry I never responded to this! I would like to set this up but prefer to use Prettier. I can handle the setup if you don't want to!

cpreston321:
I can do it! but it is opinionated! I just need to know the correct styles and format you would want for prettier! 

Thanks,
CP

ulken:
I prefer Prettier too. [Rome](https://rome.tools) is almost 100 % compatible with it as well.

cpreston321:
@natemoo-re @ulken 

I guess I would like to understand why you prefer prettier over ESLINT? ESLINT provides listing errors online and auto formats all within one tool. I think Antfu created a compelling case for it too. Read [Here](https://antfu.me/posts/why-not-prettier). I would read up on it when you have the chance. 

Thanks,
CP

ulken:
Sorry for not responding. I haven't forgotten or ignored this. Just haven't found the time/energy yet. Hopefully I'll be able to compose something later tonight. Thank you for your patience.

ulken:
# Why I use Prettier

Alright, here we go. First of all, I’m a sucker for out-of-the-box, good-defaults, zero-config tools.

Secondly, I like opinionated. Solve my problem as you know best and hide away the rest. Don’t make me choose. 

Using Prettier for the first time truly was magical. Writing your code as you wish, not having to care about formatting was almost surreal. I suppose their opinions resonating fairly well with my own beliefs plays a part in it. 

I realize a lot has happened since then and there might be better alternatives, but to me it solves the problem I have sufficiently well. Of course, performance is no downside so I’d be up for giving e.g. Rome’s formatter a go.

To me, formatting and linting fill two very different purposes, fitted for different times and places.

I want formatting to happen instantaneously when I tell it to (on save) and then get out of the way. In order to ease reading/parsing the code. 

My experience with in-line linting hasn’t been just that. It’s been intrusive, often somewhat sluggish and stalled my flow.

I’m mainly after all contributors using consistent formatting, not having to care about how they type their code. 

I realize ESLint can do formatting too, but I’m a fan of the UNIX philosophy whenever possible.

Linting has its place but I rather have it as a separate script to run on-demand (say as a prerequisite for a PR)  or as part of a CI pipeline. 

---

For the record, I have read your post but quite frankly don't find the issues brought up to be pressing, nor something I've come across frequently. But hey, that's just me 🤷

PR #27, opened 2023-02-14
feat(text): support initial value

privatenumber:
closes https://github.com/natemoo-re/clack/issues/26


natemoo-re:
Thanks for the PR! This API is quite different from what we already support, but I was able to wire up our existing `initialValue` support to the `text` component in https://github.com/natemoo-re/clack/pull/28. I'm going to close this in favor of that PR.

Really appreciate the contribution!

Issue #31, opened 2023-02-14
ERR_REQUIRE_ESM


SeverinAlexB:
Looks like the import of `ansi-regex` in `@clack/prompts/dist/index.cjs` causes my CommonJs project to crash.

```
/Users/severinbuhler/git/synonym/blocktank-instant/node_modules/@clack/prompts/dist/index.cjs:1
"use strict";const core=require("@clack/core"),color=require("picocolors"),sisteransi=require("sisteransi"),ansiRegex=require("ansi-regex"),symbol=t=>{switch(t){case"initial":case"active":return color.cyan("\u25CF");case"cancel":return color.red("\u25A0");case"error":return color.yellow("\u25B2");case"submit":return color.green("\u25CB")}},barStart="\u250C",bar="\u2502",barEnd="\u2514",text=t=>new core.TextPrompt({validate:t.validate,placeholder:t.placeholder,render(){const i=`${color.gray(bar)}
                                                                                                                      ^

Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/severinbuhler/git/synonym/blocktank-instant/node_modules/@clack/prompts/node_modules/ansi-regex/index.js from /Users/severinbuhler/git/synonym/blocktank-instant/node_modules/@clack/prompts/dist/index.cjs not supported.
Instead change the require of index.js in /Users/severinbuhler/git/synonym/blocktank-instant/node_modules/@clack/prompts/dist/index.cjs to a dynamic import() which is available in all CommonJS modules.
    at Object.<anonymous> (/Users/severinbuhler/git/synonym/blocktank-instant/node_modules/@clack/prompts/dist/index.cjs:1:119)
    at Object.<anonymous> (/Users/severinbuhler/git/synonym/blocktank-instant/dist/3_cli/index.js:26:29) {
  code: 'ERR_REQUIRE_ESM'
}

Node.js v18.12.1
```

PR #34, opened 2023-02-14
fix: `ansiRegex` package didn't have cjs imports

cpreston321:
Resolves: #31 

Description: 

It seems that the pkg didn't have correct exports for proper `cjs` import only `ESM`.

In the meantime just include it as a utility for us to use and give credit within comment.


cpreston321:
beat me to it. :)

PR #42, opened 2023-02-15
Support `isAllowEmpty` in multi-select prompt

### LINKED ISSUE #41: [Request] Allowing empty array in multi-select prompt

multi-select prompt is requires at least one option, but in some cases, we may not select anything.

How about allowing empty array in multi-select prompt?
Currently, it is possible with `validate`, but it would be nice to allow them in option.

I also considered the possibility of specifying the minimum and maximum values that can be selected with options, but I thought it would complicate things a bit.

jinmayamashita:
close #41 

It would be a little more flexible if we could use empty arrays in the multi-select prompt.
First, I think simply be used as an option (any thoughts are welcome)

### DIFF
--- a/packages/core/src/prompts/multi-select.ts
+++ b/packages/core/src/prompts/multi-select.ts
@@ -4,7 +4,7 @@ import color from 'picocolors';
 interface MultiSelectOptions<T extends { value: any }> extends PromptOptions<MultiSelectPrompt<T>> {
     options: T[];
     initialValue?: T['value'];
-    isAllowEmpty?: boolean
+    required?: boolean
 }
 export default class MultiSelectPrompt<T extends { value: any }> extends Prompt {
     options: T[];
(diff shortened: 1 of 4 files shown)


packages/core/src/prompts/multi-select.ts
```diff
@@ -4,6 +4,7 @@ import color from 'picocolors';
 interface MultiSelectOptions<T extends { value: any }> extends PromptOptions<MultiSelectPrompt<T>> {
     options: T[];
     initialValue?: T['value'];
+    isAllowEmpty?: boolean
```
natemoo-re, line 7: Would love if this was named `required` but defaulted to `true`, so this behavior would be covered by `required: false`

```suggestion
    required?: boolean
```

natemoo-re (changes requested):
Thanks for the PR! I think `required` would be a better name for this, do you mind updating the PR?

natemoo-re (approved):
Looks great, thanks for the updates!

PR #46, opened 2023-02-15
feat: `group`

cpreston321:
# POC Helper Functions

 - `definePromptGroup` - Typescript safe & easy way to define group of prompts that return JSON of all the prompts consolidated into one.
 - `definePrompt` - A way to define a prompt with the options by itself or within a group that will require a user to add additional name prop to then drive typescript to validate the options that correspond to the type
 
 ## Goal 

Make it easy to define prompts without multiple imports and that is type safe and is able to wrap into a group with additional hook that will act as the global `onCancel` to handle all cancel events within a prompt.

 ## Examples
 
 ### definePromptGroup
 
 This example is showing that it will set a global on cancel without modifying the core package that is handled by the wrapper and will run prompts in order of the array struct. This is typescript safe and will autocomplete types and options.
 
 Also note within `definePromptGroup` - `prompts: []` you **do not have** import `definePrompt` it will accept a JSON object that requires type and name. So this gives user the option.
 
 ```ts
 import { definePromptGroup, definePrompt } from "@clack/prompts"
 
 const project = await definePromptGroup({
    onCancel: ({ cancel: cancelMessage }) => {
      cancelMessage("Group Operation cancelled.");
      process.exit(0);
    },
    prompts: [
      definePrompt('text', {
        name: "path",
        options: {
          message: "Where should we create your project?",
          placeholder: "./sparkling-solid",
        },
      }),
      {
        type: 'select',
        name: 'type',
        options: {
          message: "Pick a project type.",
          options: [
            { value: "ts", label: "TypeScript" },
            { value: "js", label: "JavaScript" },
            { value: "coffee", label: "CoffeeScript", hint: "oh no" },
          ],
        },
      },
      definePrompt('multiselect', {
        name: 'tools',
        options: {
          message: "Select additional tools.",
          options: [
            { value: "prettier", label: "Prettier", hint: "recommended" },
            { value: "eslint", label: "ESLint" },
            { value: "stylelint", label: "Stylelint" },
            { value: "gh-action", label: "GitHub Action" },
          ],
        },
      })
    ]
  })
 ```

 ---

 ```ts
 const project: {
    path: string | string[];
    type: string | string[];
    tools: string | string[];
}
 ```
 
 ### definePrompt
 
 An easy way to define the prompt without an multiple imports
 
 ```ts
 import { definePrompt } from "@clack/prompts"
 
const dir = await definePrompt('text', {
        options: {
          message: "Where should we create your project?",
          placeholder: "./sparkling-solid",
        },
      })
      
if (isCancel(dir)) {
    //... logic
 }
 ```
 
 You can play around with the type safety and auto complete it's prettier sweet and enjoyable :)
 
 Thanks,
CP 🚀

### DIFF
--- a/.changeset/young-dragons-add.md
+++ b/.changeset/young-dragons-add.md
@@ -1,0 +1,5 @@
+---
+"@clack/prompts": minor
+---
+
+Add `group` construct to group many prompts together
--- a/examples/basic/index.ts
+++ b/examples/basic/index.ts
@@ -8,6 +8,8 @@ import {
   spinner,
   isCancel,
   multiselect,
+  definePromptGroup,
+  definePrompt,
   note,
 } from "@clack/prompts";
 import color from "picocolors";
@@ -20,44 +22,45 @@ async function main() {
 
   intro(`${color.bgCyan(color.black(" create-app "))}`);
 
-  const dir = await text({
-    message: "Where should we create your project?",
-    placeholder: "./sparkling-solid",
-  });
-
-  if (isCancel(dir)) {
-    cancel("Operation cancelled.");
-    process.exit(0);
-  }
-
-  const projectType = await select({
-    message: "Pick a project type.",
-    options: [
-      { value: "ts", label: "TypeScript" },
-      { value: "js", label: "JavaScript" },
-      { value: "coffee", label: "CoffeeScript", hint: "oh no" },
-    ],
-  });
-
-  if (isCancel(projectType)) {
-    cancel("Operation cancelled.");
-    process.exit(0);
-  }
-
-  const tools = await multiselect({
-    message: "Select additional tools.",
-    options: [
-      { value: "prettier", label: "Prettier", hint: "recommended" },
-      { value: "eslint", label: "ESLint" },
-      { value: "stylelint", label: "Stylelint" },
-      { value: "gh-action", label: "GitHub Action" },
-    ],
-  });
-
-  if (isCancel(tools)) {
-    cancel("Operation cancelled.");
-    process.exit(0);
-  }
+  const project = await definePromptGroup({
+    onCancel: ({ cancel: cancelMessage }) => {
+      cancelMessage("Group Operation cancelled.");
+      process.exit(0);
+    },
+    prompts: [
+      definePrompt('text', {
+        name: "path",
+        options: {
+          message: "Where should we create your project?",
+          placeholder: "./sparkling-solid",
+        },
+      }),
+      {
+        type: 'select',
+        name: 'type',
+        options: {
+          message: "Pick a project type.",
+          options: [
+            { value: "ts", label: "TypeScript" },
+            { value: "js", label: "JavaScript" },
+            { value: "coffee", label: "CoffeeScript", hint: "oh no" },
+          ],
+        },
+      },
+      definePrompt('multiselect', {
+        name: 'tools',
+        options: {
+          message: "Select additional tools.",
+          options: [
+            { value: "prettier", label: "Prettier", hint: "recommended" },
+            { value: "eslint", label: "ESLint" },
+            { value: "stylelint", label: "Stylelint" },
+            { value: "gh-action", label: "GitHub Action" },
+          ],
+        },
+      })
+    ]
+  })
 
   const install = await confirm({
     message: "Install dependencies?",
@@ -76,7 +79,7 @@ async function main() {
     s.stop("Installed via pnpm");  
   }
 
-  let nextSteps = `cd ${dir}        \n${install ? '' : 'npm install\n'}npm run dev`;
+  let nextSteps = `cd ${project.path}        \n${install ? '' : 'npm install\n'}npm run dev`;
 
   note(nextSteps, 'Next steps.');
   
--- a/examples/basic/index.ts
+++ b/examples/basic/index.ts
@@ -1,17 +1,4 @@
-import {
-  text,
-  select,
-  confirm,
-  intro,
-  outro,
-  cancel,
-  spinner,
-  isCancel,
-  multiselect,
-  definePromptGroup,
-  definePrompt,
-  note,
-} from "@clack/prompts";
+import * as p from "@clack/prompts"
 import color from "picocolors";
 import { setTimeout } from "node:timers/promises";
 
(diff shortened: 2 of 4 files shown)


natemoo-re:
Love this! Thanks for the PR.

I'm very curious if this API could be simplified. The types might be harder here, but I feel like it's less verbose. I can see the argument for abstracting away the imports, though. Thoughts?

```js
import * as p from "@clack/prompts";

const project = await p.defineGroup({
  onCancel: ({ cancel: cancelMessage }) => {
    cancelMessage("Group Operation cancelled.");
    process.exit(0);
  },
  prompts: {
    path: () =>
      p.text({
        message: "Where should we create your project?",
        placeholder: "./sparkling-solid",
      }),
    type: () =>
      p.select({
        message: "Pick a project type.",
        options: [
          { value: "ts", label: "TypeScript" },
          { value: "js", label: "JavaScript" },
          { value: "coffee", label: "CoffeeScript", hint: "oh no" },
        ],
      }),
    tools: () =>
      p.multiselect({
        message: "Select additional tools.",
        options: [
          { value: "prettier", label: "Prettier", hint: "recommended" },
          { value: "eslint", label: "ESLint" },
          { value: "stylelint", label: "Stylelint" },
          { value: "gh-action", label: "GitHub Action" },
        ],
      }),
  },
});
```

cpreston321:
> Love this! Thanks for the PR.
> 
> I'm very curious if this API could be simplified. The types might be harder here, but I feel like it's less verbose. I can see the argument for abstracting away the imports, though. Thoughts?
> 
> ```js
> import * as p from "@clack/prompts";
> 
> const project = await p.defineGroup({
>   onCancel: ({ cancel: cancelMessage }) => {
>     cancelMessage("Group Operation cancelled.");
>     process.exit(0);
>   },
>   prompts: {
>     path: () =>
>       p.text({
>         message: "Where should we create your project?",
>         placeholder: "./sparkling-solid",
>       }),
>     type: () =>
>       p.select({
>         message: "Pick a project type.",
>         options: [
>           { value: "ts", label: "TypeScript" },
>           { value: "js", label: "JavaScript" },
>           { value: "coffee", label: "CoffeeScript", hint: "oh no" },
>         ],
>       }),
>     tools: () =>
>       p.multiselect({
>         message: "Select additional tools.",
>         options: [
>           { value: "prettier", label: "Prettier", hint: "recommended" },
>           { value: "eslint", label: "ESLint" },
>           { value: "stylelint", label: "Stylelint" },
>           { value: "gh-action", label: "GitHub Action" },
>         ],
>       }),
>   },
> });
> ```

Hey! @natemoo-re 

Glad you like the idea! I think it can definitely be more simplified - my only question to this design is supporting a function to init a prompt. I guess my question is since it is a function itself, do we expect the return to be the one of the responses or was it to make where you can handle more prompts & other actions to maybe fetch data within itself?  Example: 

```ts
import * as p from "@clack/prompts";

const project = await p.defineGroup({
  onCancel: ({ cancel: cancelMessage }) => {
    cancelMessage("Group Operation cancelled.");
    process.exit(0);
  },
  prompts: {
    path: () => 
     p.text({
        message: "Where should we create your project?",
        placeholder: "./sparkling-solid",
      }),
    type: async () => {
     const res = await fetch('.....')
     const options = res.data.map((d) => ({ value: d.value : label: d.label }))
     
     return p.select({
        message: "Pick a project type.",
        options,
      })
    },
  },
});
```

I just need know since I think we would want to expect the return type to be a `prompt` to then generate the proper data for the group.

Hope to hear back 🙏🏼 

Thanks,
CP

ulken:
Cool stuff! 😎 Haven't looked closer at the code (yet), but a few high-level thoughts on what you've posted above:

Is the `define`-prefix necessary for the group? I.e. `p.defineGroup` -> `p.group`?

The callback approach allows for previous responses to be forwarded to the next prompt, like partial results. I like it. Allowing it to be async would be neat, indeed.

Can we simplify the API eve-n(ate) moore? Drop the top-level `prompts` property and simply accept properties of functions returning a prompt. Either move the `onCancel` to a second, optional options object or change the return type of the group to return results _or_ `cancel`. I kind of like the latter. Would make it consistent with the individual prompts' API as well.

In the example, why is `cancelMessage` a function?

cpreston321:
Hey @ulken based on your feedback and @natemoo-re I made some changes to simplify the process. Also to mention the `cancelMessage` function was a thought I had so you didn't have to `import { cancel } from '@clack/prompts'`. 

Here is a working version of the code!

```ts
import * as p from "@clack/prompts"

const project = await p.group({
    path: () => p.text({
      message: "Where should we create your project?",
      placeholder: "./sparkling-solid",
      validate: (value) => {
        if (!value) return "Please enter a path.";
        if (value[0] !== '.') return "Please enter an absolute path.";
      }
    }),
    type: async ({ results }) => {
      const s = p.spinner();
      const options = [
        { value: "ts", label: "TypeScript" },
        { value: "js", label: "JavaScript" },
        { value: "coffee", label: "CoffeeScript", hint: "oh no" },
      ];

      s.start("Loading project types");
      await setTimeout(2000);
      s.stop("Loaded project types");  

      return p.select({
        message: `Pick a project type within "${results.path}"`,
        initialValue: 'ts',
        options: options,
      })
    },
    tools: () => p.multiselect({
      message: "Select additional tools.",
      cursorAt: 'stylelint',
      initialValue: ['eslint', 'gh-action'],
      options: [
        { value: "prettier", label: "Prettier", hint: "recommended" },
        { value: "eslint", label: "ESLint" },
        { value: "stylelint", label: "Stylelint" },
        { value: "gh-action", label: "GitHub Action" },
      ],
    }),
    install: () => p.confirm({
      message: "Install dependencies?",
      initialValue: false
    })
  }, {
    onCancel: () => {
      p.cancel("Operation cancelled.");
      process.exit(0);
    },
  })
```

Let me know what you think.  You can also checkout and run the basic example that is working. 

```bash
pnpm start
```

Thanks 🙏

natemoo-re (approved):
🙌 This looks great, no notes!

privatenumber:
FYI since it doesn't seem to have been mentioned yet:

Unless you're only targeting ES2020, object keys are unordered and iterating over the keys can yield different results based on implementation/engine: https://stackoverflow.com/a/30919039/911407

In practice, I don't imagine this being a problem for Node.js. But JS devs that know this generally don't expect order from an object when seeing something like this:

```ts
const group = await p.group({
    name: () => p.text({ message: 'What is your name?' }),
    age: () => p.text({ message: 'What is your age?' }),
});
```

ulken:
That's a very good point, far from obvious at a first glance 👍 

However, ES2020 has been supported by Node.js since 14.X. The `target` for the lib is `ESNext` so supported by a margin.

I'd say we address this if/when people report actually getting bitten by this. In such case, I suppose we could go the React hook way and use an array instead, with destructuring to name the prompt results.

PR #51, opened 2023-02-16
Add support for neovim motion #45

### LINKED ISSUE #45: [Request] Support neovim horizontal/vertical motions
When trying to use hjkl (motion) to go through the options, nothing happens.

Get to move through options using hjkl.

**Additional context**
e.g.: When I create a new React Native app, it gives me options to select and I'm able to use hjkl to move through them.

Environment: Tmux

esau-morais:
### Description
Closes #45 

### Disclaimer
I've noticed that @natemoo-re didn't get to strict about cursor motion (horizontal/vertical), but if it is needed, I can apply it.

### DIFF
--- a/packages/core/src/prompts/prompt.ts
+++ b/packages/core/src/prompts/prompt.ts
@@ -26,19 +26,19 @@ export function isCancel(value: unknown): value is symbol {
 }
 
 function setRawMode(input: Readable, value: boolean) {
-    if ((input as typeof stdin).isTTY) (input as typeof stdin).setRawMode(value);
+  if ((input as typeof stdin).isTTY) (input as typeof stdin).setRawMode(value);
 }
 
 const keys = new Set(['up', 'down', 'left', 'right', 'space', 'enter']);
 
 export interface PromptOptions<Self extends Prompt> {
-    render(this: Omit<Self, 'prompt'>): string | void;
-    placeholder?: string;
-    initialValue?: any;
-    validate?: ((value: string) => string | void) | undefined;
-    input?: Readable;
-    output?: Writable;
-    debug?: boolean;
+  render(this: Omit<Self, 'prompt'>): string | void;
+  placeholder?: string;
+  initialValue?: any;
+  validate?: ((value: string) => string | void) | undefined;
+  input?: Readable;
+  output?: Writable;
+  debug?: boolean;
 }
 
 export type State = 'initial' | 'active' | 'cancel' | 'submit' | 'error';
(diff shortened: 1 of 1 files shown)


packages/core/src/prompts/prompt.ts
```diff
@@ -145,8 +145,24 @@ export default class Prompt {
       this.state = 'active';
     }
 
-    if (key?.name && keys.has(key.name)) {
-      this.emit('cursor', key.name);
+    switch (key?.name) {
```
ulken, line 148: Since this is the base prompt, which all other prompts extend, wouldn't this – at least, potentially others – break the `text` prompt?

There is no normal/insert modes a la `vi` for the text input, it would always be in insert mode.

packages/core/src/prompts/prompt.ts
```diff
@@ -145,8 +145,24 @@ export default class Prompt {
...
+      case 'h':
```
ulken, line 149: What do you think about adding `h, j, k, l` to `keys` and instead convert them to the corresponding cursor movement before emitting the event?

ulken (changes requested):
Took the liberty to review this. Hope you don't mind. 

1. Nice addition 👍 
2. I'd prefer for the formatting changes to be reverted. Makes the PR noisy and harder to review. If legit (not sure which rules apply), submit in a separate PR.
3. Nitpick: it's not NeoVim specific motions, rather Vi(M).

packages/core/src/prompts/prompt.ts
ulken: Nevermind, it's up to each prompt which events to subscribe to and act on. My bad.

esau-morais:
> Took the liberty to review this. Hope you don't mind.
> 
> 1. Nice addition 👍
> 2. I'd prefer the formatting changes be reverted. Makes the PR noisy and harder to review. If legit (not sure which rules apply), submit in a separate PR.
> 3. Nitpick: it's not NeoVim specific motions, rather Vi(M).

1. Thank you
2. I agree with that and I was thinking about opening a separated issue to define better documentation and also formatting/configuration
3. Yeah, you're right

packages/core/src/prompts/prompt.ts
esau-morais: Well pointed, thank you

natemoo-re:
Thanks for the PR! I just added formatting via Prettier, so the conflicts here are going to be annoying to resolve. I opened https://github.com/natemoo-re/clack/pull/56 to add your changes on top of the formatting currently on `main`.

Closing this in favor https://github.com/natemoo-re/clack/pull/56!

PR #59, opened 2023-02-17
feat: add select-key option

natemoo-re:


packages/core/src/prompts/prompt.ts
```diff
@@ -162,6 +162,9 @@ export default class Prompt {
 		if (char && (char.toLowerCase() === 'y' || char.toLowerCase() === 'n')) {
 			this.emit('confirm', char.toLowerCase() === 'y');
 		}
+		if (char) {
+			this.emit('key', char.toLowerCase());
```
Songkeys, line 166: I doubt if there's any specific use case where user wants to distinguish the cases

Songkeys: Oh I see that's the behavior we've already had. Then let's not bother it.

Songkeys:
I just checked. LGTM!

image

One unimportant thing: do you think the gray letter boxes that are vertically attached may confuse users? Would it better and capable to split them with 1px gaps so that they are horizontally linked with the same-line text in visual.

ulken:
I see your point, but AFAIK 1 px gaps are hard to achieve in a terminal. You have spaces and newlines to work with.

ulken:
Would enclosing the hotkeys in brackets (may be some better Unicode ones) more clearly visualize a key?

Songkeys:
> Would enclosing the hotkeys in brackets (may be some better Unicode ones) more clearly resemble a key?

Yes i do think so. It'd be like:

```
- [y] yes
- [n] no
- [r] regenerate
```

But the gray boxed one are cute. I'm not sure. It

---

And I think I forgot one feature here: the default selection which should be displayed in one uppercased letter. e.g.

```
- [Y] yes
- [n] no
- [r] regenerate
```

when hitting enter, it selects yes by default.

ulken:
> But the gray boxed one are cute. I'm not sure.

I agree, but I was thinking both. 

> And I think I forgot one feature here: the default selection which should be displayed in one uppercased letter. e.g.
> ```sh
> - [Y] yes
> 
> - [n] no
> 
> - [r] regenerate
> 
> ```
> 
> when hitting enter, it selects yes by default.

I think I'd prefer a higher contrast color instead.

Songkeys:
bracked letter in a gray box + default one in a higher contrasted color box.

great idea!

MendyLanda:
Hi, thanks for this PR. 

> I think I'd prefer a higher contrast color instead.

Here's what it could look like:

219868710-3a318f78-f92c-408f-a36a-f20542b93761

To my eyes, adding a default 'selected' highlight also makes it easier to differentiate from the other rows (regardless of the fact that in the example it's in the middle)

ulken:
Just to put more options on the table, here are renderings of brackets in play:

### Outer padding

Screenshot 2023-02-19 at 22 58 57

### Inner padding
Screenshot 2023-02-19 at 22 57 30

### No padding
Screenshot 2023-02-19 at 22 56 52

Thoughts?

ulken:
Yet another option: 
Screenshot 2023-02-19 at 23 13 34

I leaning towards this one. IMO, the brackets made it too clumsy, but I think the subtle hint adds value without ruining the aesthetics.

packages/core/src/prompts/select-key.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+export default class SelectKeyPrompt<T extends { value: any }> extends Prompt {
+	options: T[];
+	cursor: number = 0;
+
+	constructor(opts: SelectKeyOptions<T>) {
+		super(opts, false);
+
+		this.options = opts.options;
+		const keys = opts.options.map(v => v.value[0]);
```
ulken, line 14: If we want to restrict the value type, we could define a more narrow type:

```ts
type LowerCaseChar = 'a' | 'b' | 'c' | ... | 'z'
type Char = LowerCaseChar | UpperCase<LowerCaseChar>
type Digit = 0 | 1 | 2 | ... | 9
type AlphaNumericChar = Char | Digit
```

Of course we wouldn't support all kinds of unicode symbols, but I think it would suffice.

ulken, line 14: Since you're lowercasing the char in the event payload, shouldn't these be lowercased as well? Or use a case-insensitive check like `.test(/.../i)`.

packages/prompts/src/index.ts
```diff
@@ -191,6 +192,48 @@ export const select = <Options extends Option<Value>[], Value extends Primitive>
 	}).prompt() as Promise<Options[number]['value'] | symbol>;
 };
 
+export const selectKey = <Options extends Option<Value>[], Value extends string>(
+	opts: SelectOptions<Options, Value>
+) => {
+	const opt = (
```
ulken, line 198: I know this isn't new and follow the style rest of the code, but I prefer to spell out words. I.e. `opts` -> `options` and `opt` -> `option`. 

Any reason not to destruct the options? As in `{options, initialValue}` 

I don't find `opt` to be a descriptive name. I would prefer something along the lines `render/printOption`.

packages/prompts/src/index.ts
```diff
@@ -191,6 +192,48 @@ export const select = <Options extends Option<Value>[], Value extends Primitive>
...
+						this.options.find((opt) => opt.value === this.value)!,
+						'selected'
+					)}`;
+				case 'cancel':
+					return `${title}${color.gray(S_BAR)}  ${opt(this.options[0], 'cancelled')}\n${color.gray(
+						S_BAR
+					)}`;
+				default: {
+					return `${title}${color.cyan(S_BAR)}  ${this.options
+						.map((option, i) => opt(option))
```
ulken, line 229: `i` is unused.

ulken (changes requested):
Beautiful prompt, as always 😘 

Left some stylistic comments. Figured it'd be good to settle on some sort of standard early on.

Songkeys:
I don't have any preference with the three options above as long as there are brackets inside the box. @ulken 's proposal doesn't solve the problem that the letter in the box doesn't look like a "shortcut key" to me at the first glance. But it might be just me cuz I used to see them in brackets.

How about a poll on twitter if you couldn't decide?

ulken:
More subtle active style (with brackets back in place):

image

packages/prompts/src/index.ts
```diff
@@ -191,6 +192,48 @@ export const select = <Options extends Option<Value>[], Value extends Primitive>
...
+		option: Options[number],
+		state: 'inactive' | 'selected' | 'cancelled' = 'inactive'
+	) => {
+		const label = option.label ?? String(option.value);
+		if (state === 'selected') {
+			return `${color.dim(label)}`;
+		} else if (state === 'cancelled') {
+			return `${color.strikethrough(color.dim(label))}`;
+		}
+		return `${color.gray(color.bgWhite(color.inverse(` ${option.value} `)))} ${label}`;
```
ulken, line 208: Should `hint` be supported or not? If not, should be dropped from options.

packages/core/src/prompts/select-key.ts
ulken: Went ahead and lowercased these + lowercased in comparison.

ulken: Ignore. Irrelevant as hotkeys are generated from initial of each word.

ulken:
@natemoo-re all right, I think we're good to go here, no?

@Songkeys sorry, @natemoo-re seemed to prefer it without brackets (as do I).

ulken (approved):

natemoo-re (commented):
Thanks for jumping in @ulken! This looks great, merging.

PR #60, opened 2023-02-17
chore: add `group` to readme

cpreston321:
Updated README.md to include `group` function

### DIFF
--- a/packages/prompts/README.md
+++ b/packages/prompts/README.md
@@ -120,3 +120,36 @@ s.start('Installing via npm');
 // Do installation
 s.stop('Installed via npm');
 ```
+
+## Utilities
+
+### Grouping
+
+Grouping prompts together is a great way to keep your code organized. This accepts a JSON object with a name that can be used to reference the group later. The second argument is an optional but has a `onCancel` callback that will be called if the user cancels one of the prompts in the group.
+
+```js
+import * as p from '@clack/prompts'
+
+const group = await p.group({
+  name: () => p.text({ message: "What is your name?" }),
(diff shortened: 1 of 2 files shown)


natemoo-re (approved):
Amazing, thank you!

packages/prompts/README.md
```diff
@@ -120,3 +120,36 @@ s.start('Installing via npm');
 // Do installation
 s.stop('Installed via npm');
 ```
+
+## Utilities
+
+### Grouping
+
+Grouping prompts together is a great way to keep your code organized. This accepts a JSON object with a name that can be used to reference the group later. The second argument is an optional but has a `onCancel` callback that will be called if the user cancels one of the prompts in the group.
```
ulken, line 128: > This accepts a JSON object with a name that can be used to reference the group later.

I don't understand this part. Care to explain?

To me, there's only _one_ group of named _prompts_.

cpreston321: it was meaning to say the first argument accepts an object and for each key correlates to a prompt function that then can be referenced within the return. 

Might need to change the wording but if you have any suggestions to simplify I'll make those changes.

ulken: Right, then we're on the same page regarding intended meaning. I don't think that's what it says right now, though.

It states the name can be used to reference the _group_, not the prompt.

I'll see what I can come up with.

On a related note, I think explanations of API parameters would be better suited for more structural API docs, but that's a whole other beast.

ulken: #68

PR #61, opened 2023-02-17
feat: add password prompt to `@clack/prompts`

beerose:
I noticed that `@clack/prompts` doesn't export a password prompt, and I needed it in a project I'm currently working on 😅

Let me know if I should make any changes.

### DIFF
--- a/packages/core/src/prompts/password.ts
+++ b/packages/core/src/prompts/password.ts
@@ -10,7 +10,7 @@ export default class PasswordPrompt extends Prompt {
 	get cursor() {
 		return this._cursor;
 	}
-	private get masked() {
+	get masked() {
 		return this.value
 			.split('')
 			.map(() => this._mask)
--- a/packages/prompts/src/index.ts
+++ b/packages/prompts/src/index.ts
@@ -3,10 +3,10 @@ import {
 	ConfirmPrompt,
 	isCancel,
 	MultiSelectPrompt,
+	PasswordPrompt,
 	SelectPrompt,
 	State,
 	TextPrompt,
-	PasswordPrompt,
 } from '@clack/core';
 import isUnicodeSupported from 'is-unicode-supported';
 import color from 'picocolors';
@@ -111,7 +111,7 @@ export const password = (opts: PasswordOptions) => {
 					return `${title}${color.gray(S_BAR)}  ${color.dim(masked)}`;
 				case 'cancel':
 					return `${title}${color.gray(S_BAR)}  ${color.strikethrough(color.dim(masked ?? ''))}${
-						masked?.trim() ? '\n' + color.gray(S_BAR) : ''
+						masked ? '\n' + color.gray(S_BAR) : ''
 					}`;
 				default:
 					return `${title}${color.cyan(S_BAR)}  ${value}\n${color.cyan(S_BAR_END)}\n`;
--- a/packages/prompts/src/index.ts
+++ b/packages/prompts/src/index.ts
@@ -6,6 +6,7 @@ import {
 	SelectPrompt,
 	State,
 	TextPrompt,
+	PasswordPrompt,
 } from '@clack/core';
 import isUnicodeSupported from 'is-unicode-supported';
 import color from 'picocolors';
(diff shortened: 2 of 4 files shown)


ulken:
Cool beans!🫘 

I haven't looked at the code, nor tested it, yet, but just want to throw a thought out there:

Did you consider just adding the `mask` attribute to the `TextPrompt`?

I thought about adding this myself and was gonna explore that path first. 

Not sure just masking qualifies for a new type of prompt.

Thoughts? Any blockers/features I'm missing?

beerose:
When I used it for the first time, I kind of expected the `text` function to allow overriding how the input is printed (basically a mask), so I guess that's something we could do instead. Since there was a PasswordPromp class in `@clack/core`, I figured I could use it and create a wrapper in `@clack/prompts` (same as, for example, `TextPrompt` is used in `text`).

If we were to add `mask` to `TextPrompt` do you mean:
1) Unifying `TextPrompt` and `PasswordPrompt` classes into a single class.
2) Adding `mask` to `TextPrompt` class and `text` function and leave `PasswordPrompt` as is?
Alternatively we can add `mask` prop to `text` and use either `TextPrompt` or `PasswordPrompt` class?

ulken:
Oh, wow. I had totally missed the password prompt in core. My bad.

Well then, don't mind me. I guess @natemoo-re intended to create a separate component to go with it but hadn't come to it yet (or just missed it).

I was thinking along these lines: https://github.com/vadimdemedes/ink-text-input#mask.

ulken:
Follow-up question: shouldn't `PasswordPrompt` in core extend `TextPrompt` at least? Feels like unnecessary duplication. Or do we expect them to differ more in the future?

packages/core/src/prompts/password.ts
```diff
@@ -10,7 +10,7 @@ export default class PasswordPrompt extends Prompt {
 	get cursor() {
 		return this._cursor;
 	}
-	private get masked() {
+	get masked() {
```
ulken, line 13: Unrelated to this PR, but just curious: any particular reason we don't use `replaceAll()` here?

`return this.value.replaceAll(/./g, this._mask);`

Easier to parse/read IMO.

packages/prompts/src/index.ts
```diff
@@ -87,6 +88,38 @@ export const text = (opts: TextOptions) => {
...
+			switch (this.state) {
+				case 'error':
+					return `${title.trim()}\n${color.yellow(S_BAR)}  ${masked}\n${color.yellow(
+						S_BAR_END
+					)}  ${color.yellow(this.error)}\n`;
+				case 'submit':
+					return `${title}${color.gray(S_BAR)}  ${color.dim(masked)}`;
+				case 'cancel':
+					return `${title}${color.gray(S_BAR)}  ${color.strikethrough(color.dim(masked ?? ''))}${
+						masked?.trim() ? '\n' + color.gray(S_BAR) : ''
```
ulken, line 114: Haven't tested, but is `trim()` legit in this case? A space would be a valid password character, no?

ulken: Or rather, would this ever trim anything as spaces would be replaced by the masking character?

beerose: I followed the implementation of the `text` prompt (https://github.com/natemoo-re/clack/blob/main/packages/prompts/src/index.ts#L74) here, but I guess a space at the end is usually trimmed when you provide a password? Anyway, it won't change the underlying value, as this part of the code only prints stuff to the user's console. But let me know if I should remove the `trim()`.

packages/core/src/prompts/password.ts
ulken: Yes, sorry. They were intended for him. Should have @mentioned him.

packages/prompts/src/index.ts
ulken: > but I guess a space at the end is usually trimmed when you provide a password

I think not. I'd say that's bad practice.

> Anyway, it won't change the underlying value, as this part of the code only prints stuff to the user's console.

Yes, but what I'm saying is will this ever actually change anything? If not, I'd argue we remove it. If spaces are replaced with the mask character there will never be anything to trim.

packages/core/src/prompts/password.ts
natemoo-re: No particular reason! I'm cool with that change.

packages/prompts/src/index.ts
natemoo-re: Agreed, the trim can be removed in this case since passwords should be passed literally

natemoo-re:
> Follow-up question: shouldn't `PasswordPrompt` in core extend `TextPrompt` at least? Feels like quite a natural extension and unnecessary duplication. Or do we expect them to differ more in the future?

That seems like a very reasonable way to do it. If you want to PR that as a refactor, go for it! I'm all for anything that reduces copy + paste.

ulken:
All right, I'll do the other changes in a separate PR. So this should be good to go then?

natemoo-re (approved):
🎉 Thanks for the PR!

PR #65, opened 2023-02-20
ref: multiselect 

### LINKED ISSUE #63: [Bug] Initial value array is modified
**Environment**

- OS: macOS
- Node Version: 16.14.0
- Package:  @clack/prompts
- Package Version: 0.4.3

**Describe the bug**
I stumbled on some strange side effects, apparently the array that is passed in to `initialValue` is modified in a `multiselect` if you check in other values than the one you passed in to `initialValue`

**To Reproduce**
1. Create a multi-select prompt
2. Create an appropriate array with e.g. 3 values to use as options.
3. Create another array with 2 of the same values and use as `initialValue`.
4. Select the third value in the prompt.
5. The initialValue array will be modified.

**Expected behavior**
The array should not be mutated.

ulken:
Debugging #63, I found room for improvements of the `multiselect` prompt.

- Move options type closer to implementation
- Move required validation to styled component (we don't want styles in the headless core components)
- Remove intermediate state, always keep `value` up-to-date
- Use immutable, instead of mutating underlying, data structures 
- Remove error prone type casts
- Remove superfluous `return` handler
- Don't do `close()` on submit (already taken care of in base prompt and not always desired)
- Replace init guard
- Fix parameter type for `validate()`
- Remove non-existing cursor key case
- Remove unused parameter

Fixes #63.

### DIFF
--- a/packages/core/src/prompts/multi-select.ts
+++ b/packages/core/src/prompts/multi-select.ts
@@ -15,7 +15,8 @@ export default class MultiSelectPrompt<T extends { value: any }> extends Prompt 
 	}
 
 	private toggleValue() {
-		this.value = this.value.some(({ value }: T) => value === this._value.value)
+		const selected = this.value.some(({ value }: T) => value === this._value.value);
+		this.value = selected
 			? this.value.filter(({ value }: T) => value !== this._value.value)
 			: [...this.value, this._value];
 	}
--- a/packages/core/src/prompts/multi-select.ts
+++ b/packages/core/src/prompts/multi-select.ts
@@ -1,4 +1,3 @@
-import color from 'picocolors';
 import Prompt, { PromptOptions } from './prompt';
 
 interface MultiSelectOptions<T extends { value: any }> extends PromptOptions<MultiSelectPrompt<T>> {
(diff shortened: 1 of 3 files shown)


banjo:
Great work, code looks very clean and seems to work great when I tested it 👍

packages/core/src/prompts/multi-select.ts
```diff
@@ -10,42 +9,26 @@ interface MultiSelectOptions<T extends { value: any }> extends PromptOptions<Mul
...
 	}
 
-	private changeValue() {
-		const isValueAlreadySelected = this.selectedValues.some((value) => value === this._value.value);
-		if (isValueAlreadySelected) {
-			this.selectedValues = this.selectedValues.filter((value) => value !== this._value.value);
-		} else {
-			this.selectedValues.push(this._value.value);
-		}
+	private toggleValue() {
```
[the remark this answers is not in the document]
ulken: Valid point.

Easy to become oblivious to readability issues when you're deep into the code. Stuff makes sense at the moment, when you've already deciphered it.

I'll try out some variations.

packages/core/src/prompts/multi-select.ts
```diff
@@ -10,42 +9,26 @@ interface MultiSelectOptions<T extends { value: any }> extends PromptOptions<Mul
...
 		return this.options[this.cursor];
 	}
 
-	private changeValue() {
-		const isValueAlreadySelected = this.selectedValues.some((value) => value === this._value.value);
-		if (isValueAlreadySelected) {
-			this.selectedValues = this.selectedValues.filter((value) => value !== this._value.value);
-		} else {
-			this.selectedValues.push(this._value.value);
-		}
```
ulken, line 25: I don't quite grasp how this ever worked, since `selectedValues` is of type `T[]`, i.e. an array of _options_, not _option values_.

Hence, we're comparing options to option values. Same thing goes in UI component.

Am I missing something or did this never work or coincidentally work due to implicit type casts?

packages/core/src/prompts/multi-select.ts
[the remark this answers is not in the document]
ulken: I prefer one assignment statement so landed on your suggestion.

packages/core/src/prompts/multi-select.ts
ulken: @banjo what say you?

banjo: Yeah, I see what you mean now. I would assume that the type `T[]` for `selectedValues` is wrong from the beginning.

`this.selectedValues` comes from the `initialValue` parameter which is basically an array with strings. Or to be more exact: `initialValue?: T['value'][];`.

So even though `selectedValues` are initialized to the type `T[]`, it never actually had that value. 

A bit strange that TypeScript doesn't warn that something is wrong though?

banjo: Oh I see now. 

The problem lies in the `T` inference. 

```ts
export default class MultiSelectPrompt<T extends { value: any }>
```

It extends any class that has an object with `value: any`. By basically setting `value` to string (or whatever type `value` can be), you wouldn't be able to make that mistake as TypeScript warns again when `selectedValues` are expected to be `T[]`. 

And that part can be fixed with setting `selectedValues` to `string[]` as well.

So basically this would at least make more sense in how it was used. 

```ts
export default class MultiSelectPrompt<T extends { value: string }> extends Prompt {
	options: T[];
	cursor: number = 0;
	selectedValues: string[];

	// ...
```

Even though that really doesn't matter in this new PR 👍

natemoo-re (approved):
This looks amazing, thank you for the refactors!

packages/core/src/prompts/multi-select.ts
natemoo-re: Ah yeah nice catch, I don't think this worked as expected previously. Happy with your updates!

ulken: Yeah, I think we should go through our usages of evil `any`s and get replace them with generics of wherever possible. `any` really breaks type safety.

Issue #54, opened 2023-02-17
[Feature Request] Need  a `onCancel` property


AkaraChen:
**Is your feature request related to a problem? Please describe.**

I think have a `onCancel` property will be better.

Currently I need to call `isCanceled` in every prompts, I think it is troublesome.

**Describe the solution you'd like**

```ts
import { text, cancel } from '@clack/prompts'

const result = await text({
	/* Other options */
	onCancel() {
		cancel('Operation canceled.')
		process.exit(1)
	}
})
```

cpreston321:
Hey @AkaraChen until there is a definite solution within core then we can add these options per prompt. I'm the meantime you can wrap within a the group function and within the second param you pass onCancel like you have above. 

Example here: https://github.com/natemoo-re/clack/blob/main/examples/basic/index.ts

AkaraChen:
Thanks for such a quick reply.

Issue #37, opened 2023-02-15
Option for global onCancel


cacacaio:
**Is your feature request related to a problem? Please describe.**
It's kinda frustating to have to put a isCancel after each prompt.

**Describe the solution you'd like**
Somewhere I can define a onCancel, that always like exits the process for example.

**Describe alternatives you've considered**
Have an onCancel prop on each prompt, so that if you Ctrl + C it auto executes the function on the prop

natemoo-re:
Closed in https://github.com/natemoo-re/clack/pull/46, which will be released soon!

natemoo-re:
Released in https://github.com/natemoo-re/clack/releases/tag/%40clack%2Fprompts%400.4.0

PR #68, opened 2023-02-21
break(group): replace `onCancel` by returning `cancel`

ulken:
Might be a little controversial, but I still wanted to present the case, which I find favorable.

- Consistent API with individual prompts
- Less indirection => less confusing
- Don't wrap partial results in object => simplify destruction
- Same amount of LOC
- Rewrite docs
- Simplify implementation
- Remove nonsense `.catch()` handler
- Remove error prone `any` cast

**Note:** this would be a breaking change, so either we would have to hold off until `1.0.0` or be less semver strict (which I think is fine before `1.0.0` is released). Might want to do a `minor` at least? Changeset should be updated if so.

### DIFF
--- a/.changeset/tall-carrots-unite.md
+++ b/.changeset/tall-carrots-unite.md
@@ -1,0 +1,5 @@
+---
+"@clack/prompts": patch
+---
+
+Replace `onCancel` handler for `group()` by returning `cancel` as results instead. Partial results are no longer wrapped in an object.
--- a/examples/basic/index.ts
+++ b/examples/basic/index.ts
@@ -1,6 +1,6 @@
 import * as p from '@clack/prompts';
-import color from 'picocolors';
 import { setTimeout } from 'node:timers/promises';
+import color from 'picocolors';
 
 async function main() {
 	console.clear();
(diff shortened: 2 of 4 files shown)


natemoo-re:
Hmm, I don't disagree that this is more consistent, but I think `onCancel` might allow for a bit more flexibility? The main use case I can think of is determining which step was cancelled, which isn't automatically exposed right now but could easily be passed into the callback.

Thoughts?
/cc @cpreston321

ulken:
> Hmm, I don't disagree that this is more consistent, but I think `onCancel` might allow for a bit more flexibility? The main use case I can think of is determining which step was cancelled, which isn't automatically exposed right now but could easily be passed into the callback.

Valid point, but I don't think the flexibility is called for. What would you do with such information? 

As a user, I expect `ctrl + c` to terminate the program. Providing the possibility to print an exit message is just the right flexibility, IMO. To the user it's already obvious which step they cancelled at.

I think we should try hard to keep the API minimal and not design for hypothetical future usages no one potentially wants. But that's just my take.

cpreston321:
@ulken 
/cc @natemoo-re 

# Follow up

I am not against the simplicity and consistency, but again my goal was to create a wrapper utility function within prompts to add grouping and also by solving the case for `cancel` for multiple prompts. Then `prompts` group would mimic what core would do in the future by exposing a (callback function or a hook) to prompts that would then carry over to child package to make a easy integration in the future. 

We had similar request here on issues #37 and #54 that describe that they wanted to see the some type of `onCancel` callback within a prompt. Keeping it simple is also subjective, some people might find it easier to just work with the callback functions but then again in some cases might find it easier using the `if(isCancel(result))`..  I find the callback function is more intuitive at first glance and DX IMO but again that is just me 🤷🏼‍♂️ .

##  Limitations of implementation

- Can't check previous results filled out of the group. (only returns `symbol`) 
    - _e.g User has 10 prompts within a group. They filled out all prompts to step 8 then cancel. I want my telemetry data to be submitted of all the results to see what step they canceled on to improve prompt in the future_
    - e.g _Developer wants the user to detour to another prompt based on if a prompt was filled out within the group and matches a certain value but canceled_
    - e.g Developer wants to display clean exit message based on certain results filled out.
    
    
## Conclusion 

I think there might be alternative solution but as of right now I am not for sure what that is as of now. Hopefully we can come up with something that work in both cases and doesn't have to be one way or another. I think both have there pros and cons.

Happy to discuss more about with more developers to get there take so then we can get a good solution for the v1 release.

Thanks,
CP

natemoo-re:
Thanks for chiming in @cpreston321! I think I agree that the current API feels a bit odd, but I do think it solves some valid issues with the `Symbol('clack:cancel')` approach so I'm inclined to leave it for now.

Would love to think through some alternative APIs for cancellation. Maybe something with `AbortController` might make sense?

natemoo-re:
Going to close this for now but I'd love to continue the discussion. Will create an issue.

PR #70, opened 2023-02-21
fix(#67): multiselect return

### LINKED ISSUE #67: [Bug] Multi-Select returns options rather than values
**Environment**

- OS: Windows
- Node Version: v16.18.0
- Package: @clack/prompts
- Package Version: ^0.4.4

**Describe the bug**
As [README](https://github.com/natemoo-re/clack/tree/main/packages/prompts#multi-select) mentions, multi-select returns an array with all selected `value` props, but it returns options selected.

**To Reproduce**

https://user-images.githubusercontent.com/108878201/220267650-d41df77b-84ed-4705-a4cb-26da1a966fa6.mp4

**Expected behavior**
multi-select should return array with values selected.

ulken:
- Return `Value[]` instead of `Option[]`
- Fix type of validate parameter (somehow got lost in last PR...)

Fixes #67.

### DIFF
--- a/.changeset/dull-tools-cry.md
+++ b/.changeset/dull-tools-cry.md
@@ -1,0 +1,6 @@
+---
+"@clack/core": patch
+"@clack/prompts": patch
+---
+
+Multiselect: return `Value[]` instead of `Option[]`.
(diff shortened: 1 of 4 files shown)


natemoo-re (approved):
Thanks! I'll get this out ASAP.

PR #72, opened 2023-02-21
chore: add nvmrc file for node

banjo:
We should add an nvmrc file to keep a single source of truth. All development should occur in as similar environments as possible.


cpreston321:
Hey @banjo! 

I think @natemoo-re covers this by using [volta.sh](https://docs.volta.sh/guide/understanding#managing-your-project) basically an alternative to `nvm`.

Reference the mono-repo `package.json` here:  https://github.com/natemoo-re/clack/blob/4605547899b5cf76a3c8392dd8054b1106015c2e/package.json#L28-L30

But in the case he does want to support both I would think he would want the node versions to be the same `16.19.0`.

Thanks,
CP

banjo:
I see, I did not know that. In that case it is already covered then. I'll close this. Thank you for clarifying that!

PR #80, opened 2023-02-23
fix: hanging readline

privatenumber:
fix https://github.com/natemoo-re/clack/issues/76


natemoo-re (approved):
🙌 Wow great investigation and fix, thank you so much!

PR #82, opened 2023-02-23
feat: add log

### LINKED ISSUE #71: [Request] Allow arbitraty text as part of the chain
Currently there is not (or I couldn't find) a way to append arbitraty entries as part of the current prompt chain.
You can add new prompts, notes or spinners, but not just a single line of text without breaking the left side bar.

I've hacked a way into it by creating a fake no-time spinner like so:

```js
addText(text) {
  const sp = spinner();
  sp.start(text);
  sp.stop(text);
}

// Using it like:
intro('Welcome to this tool')
const res = await text({
  message: 'What's your name',
  placeholder: 'your name'
});

addText('My custom text as part of the prompt line')
```

It would be nice to have a straightforward way to just append a line of text, no prompt, no fancy stuff, just text that doesn't break the line with a function called something else like `message` or similar:
```js
intro('Welcome to this tool')
const res = await text({
  message: 'What's your name',
  placeholder: 'your name'
});
message('my custom text')
```

ulken:
No docs, nor examples, yet, but working implementation to play with.

Screenshot 2023-02-24 at 00 51 39

Closes #71.

### DIFF
--- a/packages/prompts/src/index.ts
+++ b/packages/prompts/src/index.ts
@@ -419,7 +419,7 @@ export type LogMessageOptions = {
 };
 export const log = {
 	message: (message = '', { symbol = color.gray(S_BAR) }: LogMessageOptions = {}) => {
-		process.stdout.write(`${color.gray(S_BAR)}\n${symbol}  ${message}\n`);
+		process.stdout.write(`${color.gray(S_BAR)}\n` + message ? `${symbol}  ${message}\n` : '');
 	},
 	info: (message: string) => {
 		log.message(message, { symbol: color.blue(S_INFO) });
--- a/packages/prompts/src/index.ts
+++ b/packages/prompts/src/index.ts
@@ -430,6 +430,10 @@ export const log = {
 	warn: (message: string) => {
 		log.message(message, { symbol: color.yellow(S_WARN) });
 	},
+	/** alias for `log.warn()`. */
+	warning: (message: string) => {
+		log.warn(message);
+	},
 	error: (message: string) => {
 		log.message(message, { symbol: color.red(S_ERROR) });
 	},
--- a/packages/prompts/src/index.ts
+++ b/packages/prompts/src/index.ts
@@ -345,19 +345,27 @@ export const multiselect = <Options extends Option<Value>[], Value extends Primi
 							i === 0 ? `${color.yellow(S_BAR_END)}  ${color.yellow(ln)}` : `   ${ln}`
 						)
 						.join('\n');
-					return `${title}${color.yellow(S_BAR)}  ${this.options
-						.map((option, i) => {
-							const selected = this.value.includes(option.value);
-							const active = i === this.cursor;
-							if (active && selected) {
-								return opt(option, 'active-selected');
-							}
-							if (selected) {
-								return opt(option, 'selected');
-							}
-							return opt(option, active ? 'active' : 'inactive');
-						})
-						.join(`\n${color.yellow(S_BAR)}  `)}\n${footer}\n`;
+					return (
+						title +
+						color.yellow(S_BAR) +
+						'  ' +
+						this.options
+							.map((option, i) => {
+								const selected = this.value.includes(option.value);
+								const active = i === this.cursor;
+								if (active && selected) {
+									return opt(option, 'active-selected');
+								}
+								if (selected) {
+									return opt(option, 'selected');
+								}
+								return opt(option, active ? 'active' : 'inactive');
+							})
+							.join(`\n${color.yellow(S_BAR)}  `) +
+						'\n' +
+						footer +
+						'\n'
+					);
 				}
 				default: {
 					return `${title}${color.cyan(S_BAR)}  ${this.options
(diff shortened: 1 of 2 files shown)


packages/prompts/src/index.ts
```diff
@@ -409,6 +414,27 @@ export const outro = (message = '') => {
...
+	message: (message = '', { symbol = color.gray(S_BAR) }: LogMessageOptions = {}) => {
+		process.stdout.write(`${color.gray(S_BAR)}\n${symbol}  ${message}\n`);
+	},
+	info: (message: string) => {
+		log.message(message, { symbol: color.blue(S_INFO) });
+	},
+	success: (message: string) => {
+		log.message(message, { symbol: color.green(S_SUCCESS) });
+	},
+	warn: (message: string) => {
```
ulken, line 430: Although `console` calls it `warn`, I'd prefer `warning` (verb vs. noun). It reads better (as in you log a warning, not a "warn") and aligns better with the rest (which are all nouns).

packages/prompts/src/index.ts
```diff
@@ -409,6 +414,27 @@ export const outro = (message = '') => {
 	process.stdout.write(`${color.gray(S_BAR)}\n${color.gray(S_BAR_END)}  ${message}\n\n`);
 };
 
+export type LogMessageOptions = {
+	symbol?: string;
+};
+export const log = {
+	message: (message = '', { symbol = color.gray(S_BAR) }: LogMessageOptions = {}) => {
```
ulken, line 421: **Note:** message is optional, so you can use this to add extra vertical spacing (newlines).

To make it more explicit, we could add another convenience function:

```ts
log.newline = () => {
  log.message();
}
```

packages/prompts/src/index.ts
natemoo-re: I see your point but I think this would trip me up since the equivalent `console.warn` is an API folks are already familiar with

packages/prompts/src/index.ts
natemoo-re: I don't think an extra method is necessary, `log.message()` works just like `console.log()` in this case!

packages/prompts/src/index.ts
```diff
@@ -409,6 +414,27 @@ export const outro = (message = '') => {
...
+		process.stdout.write(`${color.gray(S_BAR)}\n${symbol}  ${message}\n`);
```
natemoo-re, line 422: Should probably handle multiline text here

```suggestion
		const lines = message.split('\n');
		process.stdout.write(`${color.gray(S_BAR)}\n${lines.map((ln, i) => `${i === 0 ? symbol : color.gray(S_BAR)}  ${ln}`).join('\n')}\n`);
```

packages/prompts/src/index.ts
```diff
@@ -38,6 +38,11 @@ const S_CORNER_TOP_RIGHT = s('╮', '+');
 const S_CONNECT_LEFT = s('├', '+');
 const S_CORNER_BOTTOM_RIGHT = s('╯', '+');
 
+const S_INFO = s('ℹ', 'i');
+const S_SUCCESS = s('✔', '√');
+const S_WARN = s('⚠', '!');
+const S_ERROR = s('✘', 'x');
```
natemoo-re, line 44:
```suggestion
const S_INFO = s('●', '•');
const S_SUCCESS = s('◆', '*');
const S_WARN = s('▲', '!');
const S_ERROR = s('■', 'x');
```

I don't mind the traditional symbols but I think the geometric shapes align with the prompt style a bit better.

CleanShot 2023-02-24 at 17 03 28@2x

natemoo-re (changes requested):
Looks great! Few suggestions but overall really pleased with this API and implementation!

packages/prompts/src/index.ts
ulken: Fair enough. Would providing both/aliases be an option?

packages/prompts/src/index.ts
ulken: OK, should we maybe skip a newline in the empty case? I.e. just output one bar?

packages/prompts/src/index.ts
ulken: Good call!

Alternative implementation:

```ts
const [firstLine, ...lines] = message.split('\n');
process.stdout.write(`${color.gray(S_BAR)}\n${symbol}  ${firstLine}\n${lines.map((ln) => `${color.gray(S_BAR)}  ${ln}`).join('\n')}\n`);
```

packages/prompts/src/index.ts
ulken: Man, how I would kill for your designer's eye 😍

packages/prompts/src/index.ts
natemoo-re: Yeah go for it!

packages/prompts/src/index.ts
natemoo-re: I think that makes sense, yes

packages/prompts/src/index.ts
natemoo-re: Yep that's cleaner!

ulken:
Addressed your feedback. How do you feel about documentation and/or examples for this?

natemoo-re (approved):
Amazing addition, thank you so much!

Re: docs/examples—I'm hoping to knock out some docs this weekend, so I can tackle this.

PR #86, opened 2023-02-26
fix(#44) complex values

### LINKED ISSUE #44: [Request] Support non-string select / multiselect `value`s
The TypeScript types for `select` and `multiselect` options enforce that `value` fields passed extend `Readonly<string>`:
https://github.com/natemoo-re/clack/blob/d366a1f6e2339b5b97d0b9f42a8bdfd86333c2ba/packages/prompts/src/index.ts#L118-L120

This can be a bit cumbersome to use with non-string values, as it requires stringifying to pass to `select` + parsing the result returned back.

Ideally (assuming no technical blockers + context I'm missing), `select` and `multiselect` could support any type of value. (perhaps naively) it looks like this could be as simple as changing the code block above to support a looser `Value` generic.

Stringifying + parsing as described above, though this isn't always feasible for more complex types.

**Additional context**
Interestingly, from my (relatively basic) testing, this only seems to be a type error. Passing in `Date` values works as expected - are there other types of values which may break this?

If the fix is as simple as I'm imagining, I'd be happy to open a PR and take a stab at this! However, I'm still fairly early in my testing of this library so I may be missing some important context of why this limitation exists :)

ulken:
Select prompts (incl. multi + group-multi):

- simplify option value types
- support complex value types

Screenshot 2023-02-26 at 01 06 11

Fixes #44 (for realz). 

**Note:** `selectKey` is not supported since it requires `value` to be `string`.

### DIFF
--- a/packages/prompts/src/index.ts
+++ b/packages/prompts/src/index.ts
@@ -177,16 +177,13 @@ interface Option<Value extends Primitive> {
 export interface SelectOptions<Options extends Option<Value>[], Value extends Primitive> {
 	message: string;
 	options: Options;
-	initialValue?: Options[number]['value'];
+	initialValue?: Value;
 }
 
 export const select = <Options extends Option<Value>[], Value extends Primitive>(
 	opts: SelectOptions<Options, Value>
 ) => {
-	const opt = (
-		option: Options[number],
-		state: 'inactive' | 'active' | 'selected' | 'cancelled'
-	) => {
+	const opt = (option: Option<Value>, state: 'inactive' | 'active' | 'selected' | 'cancelled') => {
 		const label = option.label ?? String(option.value);
 		if (state === 'active') {
 			return `${color.green(S_RADIO_ACTIVE)} ${label} ${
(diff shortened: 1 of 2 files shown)


packages/prompts/src/index.ts
```diff
@@ -169,24 +169,21 @@ export const confirm = (opts: ConfirmOptions) => {
...
 type Primitive = Readonly<string | boolean | number>;
-interface Option<Value extends Primitive> {
-	value: Value;
-	label?: string;
-	hint?: string;
-}
-export interface SelectOptions<Options extends Option<Value>[], Value extends Primitive> {
+
+type Option<Value> = Value extends Primitive
+	? { value: Value; label?: string; hint?: string }
```
ulken, line 174: Suppose we could extract aliases for these, if it would make easier to decipher violations?

```ts 
type PrimitiveOption<Value extends Primitive> = { value: Value; label?: string; hint?: string>

type ComplexOption<Value> = { value: Value; label: string; hint?: string>
```

natemoo-re (approved):
🙌 Thanks, nice fix!

packages/prompts/src/index.ts
ulken: @natemoo-re thoughts?

PR #89, opened 2023-03-02
chore: add editor files

privatenumber:
Added:
- `.editorconfig` to configure Editors
- `.gitattributes` to configure Git
- `.vscode` to tell VSCode to use the local TypeScript version instead of the global

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+root = true
+
+[*]
+indent_style = tab
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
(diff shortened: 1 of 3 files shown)


ulken (approved):
LGTM 👍

privatenumber:
Do you have merge permissions?

With these PRs, I'm making my towards fixing a type bug in clack so I'd like to move faster.

ulken:
I do. Although @natemoo-re hasn't given me any explicit directions of what to do or not, I see little harm in merging these. Let's merge.

Worst case, he'll have to revert and be more explicit.

PR #90, opened 2023-03-02
ci: setup type-checking on PRs

privatenumber:
Since unbuild doesn't type-check, there's been quite a few type-errors that got in.

Now CI will run TypeScript to make sure there are no errors.


privatenumber:
While this PR is failing, I can make separate PRs to `main` to fix the issues and pull them back in here until type-checking passes.

ulken (approved):
This has disturbed me as well, so all for 👍

Minor: but the title/commit isn't 💯 true as it runs on pushes to `main` too.

natemoo-re (approved):
Makes total sense, thanks for the PR!

Issue #92, opened 2023-03-02
Async validation


SeverinAlexB:
**Is your feature request related to a problem? Please describe.**
I ask the user for an account id. I like to validate this id by checking it against the db which I can't do sync.

**Describe the solution you'd like**
Make `validate` compatible with `async`.

**Describe alternatives you've considered**
Currently I load all the accounts before the prompt so I can validate the id sync. This doesn't scale though.

**Additional context**
That's it :)

ulken:
Follow-up to keep you in the loop: I started looking into this, but quickly realized we need to provide feedback to the user while waiting. 

Not entirely obvious how that would fit into the current implementation flow. Asked for input, but haven't gotten any, yet.

SeverinAlexB:
Thx. I see your point. It is already worth having it without a spinner IMO. Most validation flows should be quick (db calls are in the ms). But I understand in case you want to have it perfect.

ulken:
Progress has been made. Have some rough edges to smooth out.

stordahl:
Would love to see this ship 🙏

dylang:
I appreciate your prioritization on human interaction, but for my particular usage it would be helpful to have the ability to use async validators, even with no visual progress indicator, rather than no async support at all.

PR #94, opened 2023-03-02
async validate

ulken:
Took a stab at supporting async validation. 

Currently, the UX is suffering, though. The program seemingly hangs, without any feedback to the user. We should probably kick off a spinner or similar.

Potentially by emitting a `validate` event from core to allow styled prompts to take appropriate action? Or introducing a new state and doing an intermediate render?

Related to #92

### DIFF
--- a/examples/basic/index.ts
+++ b/examples/basic/index.ts
@@ -15,7 +15,8 @@ async function main() {
 				p.text({
 					message: 'Where should we create your project?',
 					placeholder: './sparkling-solid',
-					validate: (value) => {
+					validate: async (value) => {
+						await setTimeout(299);
 						if (!value) return 'Please enter a path.';
 						if (value[0] !== '.') return 'Please enter a relative path.';
 					},
--- a/examples/basic/index.ts
+++ b/examples/basic/index.ts
@@ -15,8 +15,7 @@ async function main() {
 				p.text({
 					message: 'Where should we create your project?',
 					placeholder: './sparkling-solid',
-					validate: async (value) => {
-						await setTimeout(299);
+					validate: (value) => {
 						if (!value) return 'Please enter a path.';
 						if (value[0] !== '.') return 'Please enter a relative path.';
 					},
--- a/packages/core/src/prompts/prompt.ts
+++ b/packages/core/src/prompts/prompt.ts
@@ -38,11 +38,15 @@ const aliases = new Map([
 ]);
 const keys = new Set(['up', 'down', 'left', 'right', 'space', 'enter']);
 
+export interface Validator<Value = any> {
+	(value: Value): string | void | Promise<string | void>;
+}
+
 export interface PromptOptions<Self extends Prompt> {
 	render(this: Omit<Self, 'prompt'>): string | void;
 	placeholder?: string;
 	initialValue?: any;
-	validate?: ((value: any) => string | void) | undefined;
+	validate?: Validator;
 	input?: Readable;
 	output?: Writable;
 	debug?: boolean;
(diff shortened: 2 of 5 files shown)


ulken:
@natemoo-re thoughts?

natemoo-re:
This looks good! I pushed a change to set a `validate` state, but it only triggers after 300ms so the UI doesn't flash a temporary loading state if the validation is pretty fast.

Also adds a `validate` state, which can be triggered during a long async validation.

natemoo-re:
Feel free to merge if you feel comfortable with this!

ulken:
Nice solution! I'll make sure to try it out and merge if appeased 😉

examples/basic/index.ts
```diff
@@ -15,7 +15,8 @@ async function main() {
 				p.text({
 					message: 'Where should we create your project?',
 					placeholder: './sparkling-solid',
-					validate: (value) => {
+					validate: async (value) => {
+						await setTimeout(299);
```
ulken, line 19: Curious as to why you've added the timeout here when the delay is below 300 ms? Meaning it won't show, right? Why not leave it out all together then?

packages/core/src/prompts/prompt.ts
```diff
@@ -168,7 +173,17 @@ export default class Prompt {
...
 		if (key?.name === 'return') {
 			if (this.opts.validate) {
-				const problem = this.opts.validate(this.value);
+				this.state = 'validate';
+				let problem = this.opts.validate(this.value);
+				// Only trigger validation state after 300ms.
+				// If problem resolves first, render will be cancelled.
+				await Promise.race([
+					problem,
+					setTimeout(300).then(() => {
```
ulken, line 182: Magic number? Constant with descriptive name?

Just curious, but isn't 400 the typical breakpoint in these scenarios?

examples/basic/index.ts
natemoo-re: Ooh yeah that was just for testing purposes, can remove!

packages/core/src/prompts/prompt.ts
natemoo-re: Yeah, magic number. Moving to a constant with descriptive name is probably better! 400 works for me 👍🏻

ulken:
The password prompt seems to be broken now (compared to the main branch — even after handling the `validate` state). When submitting the masked value is seemingly cleared.

ulken:
It seems like `Promise.race()` doesn't work as you expect with `setTimeout()`. It's not cancelled (thus always resolved and rendered). You would need to pass a `signal` (AbortController) to cancel the timeout.

packages/core/src/prompts/prompt.ts
```diff
@@ -172,7 +177,17 @@ export default class Prompt {
 
 		if (key?.name === 'return') {
 			if (this.opts.validate) {
-				const problem = this.opts.validate(this.value);
+				this.state = 'validate';
```
ulken, line 180: When the `race`/`timeout` is sorted, shouldn't this move closer to the `render`? No need to set it if never rendered?

ulken:
Would have to use something like: https://stackoverflow.com/a/39852372

ulken:
Gah, for unknown reason, the password prompt doesn't like `onKeypress` to be async.

Simply add an `await Promise.resolve()` in there and the input is cleared when pressing Enter.

Not sure if readline doesn't like async or something.

ulken:
Pushed fix for the timeout issue, but have yet to resolve the password prompt not working. I can't understand why it works for text but not password...

Mist3rBru:
@natemoo-re @ulken @cpreston321 I looked at this, and seems the password prompt was losing the prompt context during async validation, I fixed it on this commit [d258379](https://github.com/Mist3rBru/clack/commit/d2583796a50a7d5f08401588c870254e26fead51), please check it out!

negativems:
How is the progress on this?

bdrtsky:
Too bad it's unsupported yet

43081j:
Cleaning up old PRs, so closing this.

The issue remains open, so we will get to it at some point 👍  I have a branch in progress

PR #95, opened 2023-03-03
ci: build on prepack

privatenumber:
As per https://github.com/natemoo-re/clack/commit/6233de11802733479193f37ead3f62f801c1b12c


ulken (approved):
👍

PR #97, opened 2023-03-05
Fix line duplication bug with proper line wrapping

### LINKED ISSUE #87: [Bug] Duplicate lines
**Environment**

- OS: macOS
- Node Version: v19.1.0
- Package: @clack/prompts
- Package Version: v0.3.0

**Describe the bug**

The terminal duplicates options in a select when moving arrows up and down if the window is too small and the lines try to wrap. More of a rendering bug, because if you expand the terminal, it seems to start working normally, but the duplicates stay and aren't selectable.

**To Reproduce**

https://stackblitz.com/edit/node-omu3tm?file=prompts.js

-  Put several options in a select, then shrink the terminal width so an option splits into multiple lines.

**Expected behavior**

The select should behave normally, moving arrows up and down should not duplicate everything. This seems to be a limitation of the terminal itself, perhaps truncation and an ellipsis of the long options would help.

natemoo-re:
Closes #87.

- Bundles `wrap-ansi`. Kind of a bummer, but it's battle-tested and works very well. Ideally we'll continue exploring our own, smaller version on the [`feat/wrap`](https://github.com/natemoo-re/clack/tree/feat/wrap) branch.
- Automatically wraps prompts to properly calculate visual line offset for each frame
- Also re-wraps the previous frame in case the columns have changed. This allows the line calculation to stay up-to-date.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -55,5 +55,8 @@
   "dependencies": {
     "picocolors": "^1.0.0",
     "sisteransi": "^1.0.5"
+  },
+  "devDependencies": {
+    "wrap-ansi": "^8.1.0"
   }
 }
--- a/packages/core/src/prompts/prompt.ts
+++ b/packages/core/src/prompts/prompt.ts
@@ -5,6 +5,7 @@ import readline from 'node:readline';
 import { Readable, Writable } from 'node:stream';
 import { WriteStream } from 'node:tty';
 import { cursor, erase } from 'sisteransi';
+import wrap from 'wrap-ansi';
 
 function diffLines(a: string, b: string) {
 	if (a === b) return;
@@ -104,17 +105,20 @@ export default class Prompt {
 
 		this.input.on('keypress', this.onKeypress);
 		setRawMode(this.input, true);
+		this.output.on('resize', this.render);
 
 		this.render();
 
 		return new Promise<string | symbol>((resolve, reject) => {
 			this.once('submit', () => {
 				this.output.write(cursor.show);
+				this.output.off('resize', this.render);
 				setRawMode(this.input, false);
 				resolve(this.value);
 			});
 			this.once('cancel', () => {
 				this.output.write(cursor.show);
+				this.output.off('resize', this.render);
 				setRawMode(this.input, false);
 				resolve(cancel);
 			});
@@ -203,13 +207,13 @@ export default class Prompt {
 
 	// TODO: handle wrapping
 	private restoreCursor() {
-		const lines = this._prevFrame.split('\n').length - 1;
+		const lines = wrap(this._prevFrame, process.stdout.columns).split('\n').length - 1;
 		this.output.write(cursor.move(-999, lines * -1));
 	}
 
 	private _prevFrame = '';
 	private render() {
-		const frame = this._render(this) ?? '';
+		const frame = wrap(this._render(this) ?? '', process.stdout.columns);
 		if (frame === this._prevFrame) return;
 
 		if (this.state === 'initial') {
--- a/.changeset/good-islands-provide.md
+++ b/.changeset/good-islands-provide.md
@@ -1,0 +1,6 @@
+---
+'@clack/core': patch
+'@clack/prompts': patch
+---
+
+Fix line duplication bug by automatically wrapping prompts to `process.stdout.columns`
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -50,9 +50,12 @@ importers:
     specifiers:
       picocolors: ^1.0.0
       sisteransi: ^1.0.5
+      wrap-ansi: ^8.1.0
     dependencies:
       picocolors: 1.0.0
       sisteransi: 1.0.5
+    devDependencies:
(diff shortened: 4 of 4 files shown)


packages/core/package.json
```diff
@@ -55,5 +55,8 @@
   "dependencies": {
     "picocolors": "^1.0.0",
     "sisteransi": "^1.0.5"
+  },
+  "devDependencies": {
```
ulken, line 59: Why a dev dep?

packages/core/src/prompts/prompt.ts
```diff
@@ -104,17 +105,20 @@ export default class Prompt {
...
 
 		return new Promise<string | symbol>((resolve, reject) => {
 			this.once('submit', () => {
 				this.output.write(cursor.show);
+				this.output.off('resize', this.render);
 				setRawMode(this.input, false);
 				resolve(this.value);
 			});
 			this.once('cancel', () => {
 				this.output.write(cursor.show);
```
ulken, line 120: Time to refactor into a clean up function of sorts? (For both promise branches, I mean)

ulken (commented):
Sweetz! 

Let me know if you want me to look at/discuss something in the custom implementation. Would it be possible to document what works and not?

PS. Good to see you back in the game 😉 It's been awfully quiet the last week 😁 I took the liberty to merge a few minor PRs. There were some question marks directed towards you, so if possible it would be good to sort out.

ulken:
@natemoo-re the release build seem to have failed. Did we (I) break something by merging https://github.com/natemoo-re/clack/pull/91 or https://github.com/natemoo-re/clack/pull/95?

The latter seems more likely if so. Removing the `build` command seems suspicious. Is `prepack` not run when/as expected after all?

There were questions around your previous change there so probably related.

ulken:
// cc @privatenumber

ulken:
Potentially related and maybe what @natemoo-re followed?

https://github.com/changesets/changesets/issues/837#issuecomment-1222206876

ulken:
Still don't understand why `prepack` is not called, though.

`pnpm publish` should trigger it if I understand this correct? https://pnpm.io/cli/publish

and `changesets` seem to use it if installed?

https://github.com/changesets/changesets/blob/5c53bff0efeb5bbd72976b79aefe9d5805038ebd/packages/cli/src/commands/publish/npm-utils.ts#L156

ulken:
Should probably revert in the meantime?

natemoo-re:
> @natemoo-re the release build seem to have failed. Did we (I) break something by merging https://github.com/natemoo-re/clack/pull/91 or https://github.com/natemoo-re/clack/pull/95?
> 
> The latter seems more likely if so. Removing the `build` command seems suspicious. Is `prepack` not run when/as expected after all?
> 
> There were questions around your previous change there so probably related. 

Ah whoops! I pressed merge and then stepped out. My bad! I'll have to look into what happened here.

natemoo-re:
I think https://github.com/natemoo-re/clack/commit/a7e6a5a6bcc7e8aa86de29b82f5378fee4b133fe should fix it. I haven't used `prepack` before, maybe `changesets` doesn't automatically call it...?

ulken:
> > @natemoo-re the release build seem to have failed. Did we (I) break something by merging https://github.com/natemoo-re/clack/pull/91 or https://github.com/natemoo-re/clack/pull/95?
> > 
> > The latter seems more likely if so. Removing the `build` command seems suspicious. Is `prepack` not run when/as expected after all?
> > 
> > There were questions around your previous change there so probably related. 
> 
> Ah whoops! I pressed merge and then stepped out. My bad! I'll have to look into what happened here.

If anyone should apologize, it's me.

ulken:
> I think https://github.com/natemoo-re/clack/commit/a7e6a5a6bcc7e8aa86de29b82f5378fee4b133fe should fix it. I haven't used `prepack` before, maybe `changesets` doesn't automatically call it...?

Right, but `prepack` is one of many [lifecycle scripts](https://docs.npmjs.com/cli/v9/using-npm/scripts#life-cycle-scripts) called by the package manager. `pnpm` [supports them](https://pnpm.io/cli/publish) too.

ulken:
And judging by the code it seems to invoke the underlying package manager.

ulken:
Let's leave it at this for now. @privatenumber might want to investigate further.

Might be good to add a comment in the code?

natemoo-re:
> If anyone should apologize, it's me.

No worries! Hard to know if CI changes are going to work until they run... This one would have tripped me up as well.

> Right, but `prepack` is one of many [lifecycle scripts](https://docs.npmjs.com/cli/v9/using-npm/scripts#life-cycle-scripts) called by the package manager. `pnpm` [supports them](https://pnpm.io/cli/publish) too.

Yeah I would expect this to work! I'm not sure why `prepack` _wouldn't_ work, but better safe than sorry I guess?

natemoo-re:
I'll add a comment!

ulken:
TIL: https://github.com/pnpm/pnpm/issues/2891

`prepack` is not user-defined, though, but I wasn't aware of this difference between `npm` and `yarn`/`pnpm`.

PR #107, opened 2023-03-14
Prettify group return type

chrissantamaria:
👋 tiny enhancement PR - I noticed that the return type exposed to consumers of `select` is wrapped by `PromptGroupAwaitedReturn`:

Screenshot 2023-03-14 at 2 31 02 PM

This has two undesirable consequences:
- `PromptGroupAwaitedReturn` is an abstraction which consumers shouldn't worry about - functionally, they get back a bare object so the types should reflect that
- `PromptGroupAwaitedReturn` implicitly strips `symbol` types from keys - ideally, consumers shouldn't worry about this either and instead just get back the types representing returned values

To make this more clear, this PR wraps the return in a `Prettify` helper type to make the exposed type a bare object:

Screenshot 2023-03-14 at 2 30 32 PM

This should be a functional noop, but let me know if you notice any undesirable consequences! This pattern is also used [a fair amount](https://sourcegraph.com/search?q=context%3Aglobal+lang%3ATypeScript+%22%26+%7B%7D%22&patternType=standard&sm=1&groupBy=repo) in other TypeScript projects.

### DIFF
--- a/packages/prompts/src/index.ts
+++ b/packages/prompts/src/index.ts
@@ -58,6 +58,8 @@ const symbol = (state: State) => {
 	}
 };
 
+type Prettify<T> = { [K in keyof T]: T[K] } & {};
+
 export interface TextOptions {
 	message: string;
 	placeholder?: string;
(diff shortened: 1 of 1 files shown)


ulken:
Thank you for your detailed PR, but someone beat you to it 😉: https://github.com/natemoo-re/clack/pull/105#issuecomment-1464009417

ulken:
Since the former has been merged, I'm gonna go ahead and close this.

chrissantamaria:
hah, funny timing - had been thinking about that for a while but never got around to making a PR 😛

PR #113, opened 2023-03-29
feat(#106): `spinner` hook to handle process exit

### LINKED ISSUE #106: [Request] Spinner to handle cleaning up on process exit

When the logic between `spinner.start()` & `spinner.stop()` throws an error and terminates the process, it seems to cause issues with the terminal cursor:

https://github.com/Nutlope/aicommits/pull/154/

So to work around this, the logic block needs to be wrapped in a `try-finally` to call `spinner.stop()`.

An exit hook inside `spinner` to cleanup the `readline` prompt if the process is about to exit:

```js
process.on('exit', () => {
    if (spinnerActive) {
    	// terminate readline
    }
})
```

Delegate `try-finally` wrappers to user-land, but it's not very ergonomic to do this every time `spinner` is used.

**Additional context**

Mist3rBru:
Closes  #106 

It adds a hook to `spinner`, that prints a default message and avoid terminal conflicts, it triggers on:
- Code exception
- Promise rejection
- Ctrl + C
- Kill process
- Process shutdown

### DIFF
--- a/packages/prompts/src/index.ts
+++ b/packages/prompts/src/index.ts
@@ -657,7 +657,7 @@ export interface PromptGroupOptions<T> {
 	 * Control how the group can be canceled
 	 * if one of the prompts is canceled.
 	 */
-	onCancel?: (opts: { results: Partial<PromptGroupAwaitedReturn<T>> }) => void;
+	onCancel?: (opts: { results: Prettify<Partial<PromptGroupAwaitedReturn<T>>> }) => void;
 }
 
 type Prettify<T> = {
--- a/packages/prompts/src/index.ts
+++ b/packages/prompts/src/index.ts
@@ -654,15 +654,19 @@ export type PromptGroupAwaitedReturn<T> = {
 
 export interface PromptGroupOptions<T> {
 	/**
-	 * Control how the group can be canceld
-	 * if one of the prompts is canceld.
+	 * Control how the group can be canceled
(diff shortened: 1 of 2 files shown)


Mist3rBru:
Code with `try-catch` logic, will still be able to call `spinner.stop()` and print their message

cpreston321:
Hey @Mist3rBru 

While testing I did see this: 

image

If you want to test yourself: `cd examples/basic && pnpm start`

The code looks good but it seems that maybe something messed up when merge. I just did a `CRTL+C` when installing deps  was showing.

Thanks,
CP 🚀

cpreston321 (changes requested):
I left some comments above based on testing.

packages/prompts/src/index.ts
```diff
@@ -635,44 +635,71 @@ export const log = {
...
+		clearInterval(loop);
+		const step =
+			code === 0
+				? color.green(S_STEP_SUBMIT)
+				: code === 1
+				? color.red(S_STEP_CANCEL)
+				: color.red(S_STEP_ERROR);
+		process.stdout.write(cursor.move(-999, 0));
+		process.stdout.write(erase.down(1));
+		process.stdout.write(`${step}  ${message}\n`);
```
cpreston321, line 677:
```suggestion
		process.stdout.write(`${step}  ${_message}\n`);
```

PR #119, opened 2023-04-17
fix: note title correctly evaluates title length with ansi

SimeonC:
The bug is caused when a title has ansi characters. In my case it was when I used chalk in both the title and note which caused the below screenshot.

Before this change;
CleanShot 2023-04-17 at 14 11 42

After change;
CleanShot 2023-04-17 at 14 11 59


SimeonC:
@natemoo-re What am I missing to get a review/approval on this PR?

ConProgramming:
@natemoo-re would love to have this

cpreston321 (approved):
Thanks 🙏🏼  but LGTM

PR #124, opened 2023-05-26
feat: implement filter list

alexzhang1030:
Quick preview:

preview

Currently stage: *draft*

Need some feedback and suggestions:

- Do we need to separate the filter mode to avoid key conflicts caused by movement (hjkl) and selection (space) in normal mode?

usage:

```ts
p.select({
  options: [{ value: 'basic', label: 'Basic' }],
  message: 'Select an example to run.',
  enableFilter: true, // enable filterable
});
```

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -57,6 +57,7 @@
     "sisteransi": "^1.0.5"
   },
   "devDependencies": {
-    "wrap-ansi": "^8.1.0"
+    "wrap-ansi": "^8.1.0",
+		"fzf": "^0.5.2"
   }
 }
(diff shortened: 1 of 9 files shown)


packages/core/src/prompts/select.ts
```diff
@@ -19,7 +22,7 @@ export default class SelectPrompt<T extends { value: any }> extends Prompt {
 	constructor(opts: SelectOptions<T>) {
 		super(opts, false);
 
-		this.options = opts.options;
+		this.originalOptions = this.options = opts.options;
```
vsn4ik, line 25: Maybe:
```suggestion
		this.originalOptions = opts.options;
```

packages/core/package.json
```diff
@@ -57,6 +57,7 @@
     "sisteransi": "^1.0.5"
   },
   "devDependencies": {
-    "wrap-ansi": "^8.1.0"
+    "wrap-ansi": "^8.1.0",
+		"fzf": "^0.5.2"
```
vsn4ik, line 61:
```suggestion
	"fzf": "^0.5.2"
```

alexzhang1030: Ye, my bad.

vsn4ik (approved):

PR #129, opened 2023-06-21
feat(@clack/prompts): add `maxItems` option to select prompt

lforst:
This PR adds a `maxItems` option to `@clack/prompts`'s select prompt that allows to limit the number of shown options at any given time.

When using this option, any list of options that exceeds `maxItems` will instead show a scrollable list. The "scrollability" is denoted by ellipses ("...").

I am not opinionated about the name `maxItems` for the option. We could also name it `maxItemsShown` or something similar.

Ref https://github.com/getsentry/sentry-wizard/issues/286
Resolves https://github.com/natemoo-re/clack/issues/118

---

Example:

https://github.com/natemoo-re/clack/assets/8118419/72577604-ac18-4538-ba71-0fd19159f50b

### DIFF
--- a/packages/prompts/src/index.ts
+++ b/packages/prompts/src/index.ts
@@ -178,6 +178,7 @@ export interface SelectOptions<Options extends Option<Value>[], Value> {
 	message: string;
 	options: Options;
 	initialValue?: Value;
+	maxItems?: number;
 }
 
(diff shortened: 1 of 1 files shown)


peterroe:
I need this feature urgently!!!

cpreston321 (approved):

PR #139, opened 2023-08-11
feat(@clack/prompts): add prompt `workflow`

### LINKED ISSUE #109: [Request] logic diagram
### Is your feature request related to a problem?
included that scenario

and including return an particular step if ...

### Describe the solution you'd like
like prompt type but with chain question

### Additional context
provide a FAQ on yes/no like

### LINKED ISSUE #131: [Bug] can't infer the input type when using results in p.group
**Environment**

- OS: MacOS
- Node Version: 18.16.0
- Package: @clack/prompts
- Package Version: 0.6.3

**Describe the bug**

just copy the example code, and check the `type`:

https://github.com/natemoo-re/clack/blob/main/examples/basic/index.ts#L31

image

**To Reproduce**

just use the basic example code

**Expected behavior**

project.type should be infered to type: `string`

**Additional Information**

orochaa:
Since Typescript has shown a limitation on infer `group` results, discussed in @chrissantamaria's [comment on PR#102]( https://github.com/natemoo-re/clack/pull/102#issuecomment-1465068829), and Issue #131. This functionality presents an alternative for users who favor the builder pattern when dealing with multiple prompts in sequence, serving as either a workaround or an additional choice, while avoid a breaking change.

This PR includes:
- [x] Code
- [x] Docs
- [x] Changeset
- [x] Example

I am open to discussions on comment's section!

Closes #131
Closes #109

### DIFF
--- a/packages/prompts/README.md
+++ b/packages/prompts/README.md
@@ -159,6 +159,72 @@ const group = await p.group(
 console.log(group.name, group.age, group.color);
 ```
 
+### Workflow
+
+Works just like `group` but infer types way better and treats your group like a workflow, allowing you to create conditional steps (forks) along the process.
+
+```js
+import * as p from '@clack/prompts';
+
+const results = await p
+  .workflow()
+  .step('name', () => p.text({ message: 'What is your package name?' }))
+  .step('type', () =>
+    p.select({
+      message: `Pick a project type:`,
+      initialValue: 'ts',
+      maxItems: 5,
+      options: [
+        { value: 'ts', label: 'TypeScript' },
+        { value: 'js', label: 'JavaScript' },
+        { value: 'rust', label: 'Rust' },
+        { value: 'go', label: 'Go' },
+        { value: 'python', label: 'Python' },
+        { value: 'coffee', label: 'CoffeeScript', hint: 'oh no' },
+      ],
+    })
+  )
+  .step('install', () =>
+    p.confirm({
+      message: 'Install dependencies?',
+      initialValue: false,
+    })
+  )
+  .step('fork', ({ results }) => {
+    if (results.install === true) {
+      return p
+        .workflow()
+        .step('package', () =>
+          p.select({
+            message: 'Pick a package manager:',
+            initialValue: 'pnpm',
+            options: [
+              {
+                label: 'npm',
+                value: 'npm',
+              },
+              {
+                label: 'yarn',
+                value: 'yarn',
+              },
+              {
+                label: 'pnpm',
+                value: 'pnpm',
+              },
+            ],
+          })
+        )
+        .run();
+    }
+  })
+  .onCancel(() => {
+    p.cancel('Workflow canceled');
+    process.exit(0);
+  })
+  .run();
+console.log(results);
+```
+
 ### Tasks
 
 Execute multiple tasks in spinners.
--- a/examples/basic/package.json
+++ b/examples/basic/package.json
@@ -13,7 +13,8 @@
     "stream": "jiti ./stream.ts",
     "spinner": "jiti ./spinner.ts",
     "spinner-ci": "npx cross-env CI=\"true\" jiti ./spinner-ci.ts",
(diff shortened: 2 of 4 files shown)


.changeset/config.json
```diff
@@ -7,5 +7,5 @@
   "access": "public",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
-  "ignore": ["@example/*"]
```
cpreston321, line 10: @Mist3rBru we include this because we don't publish any packages within `@examples/` and all examples have namespaces.

Mist3rBru: I noticed that all examples would copy-paste the same template, so I group then on examples folder and tested Changeset, it still ignores the examples, if you prefer to keep that I can revert it.

cpreston321: I like this actually, as I was thinking of it more. I think we could leverage this to be used in within https://clack.cc to switch between different examples.

Mist3rBru: That would be flat out nice!

cpreston321:
## Initial thoughts: 

@Mist3rBru Great Work 🙏🏼 , 

I actually kinda of like this.. This could be a great addition. I'm just not for sure on the namespace `builder`. It almost sounds to me like a `workflow`. What are your thoughts?  

### Additional thoughts

I also think, within the next couple of releases -- I want to focus on knocking out a few bugs reported, then exposing each individual `core` prompt interface types to make it agnostic vs duplicating types that are already exist in `core`. This way someone can easily build on top of core as an agnostic layer like what we do for `@clack/prompts`. Then we can add another layer for themes etc. 

Thanks,
CP 🚀

packages/prompts/README.md
```diff
@@ -156,3 +156,32 @@ const group = await p.group(
...
+    p.multiselect({
+      message: `What is your favorite color ${results.name}?`,
+      options: [
+        { value: 'red', label: 'Red' },
+        { value: 'green', label: 'Green' },
+        { value: 'blue', label: 'Blue' },
+      ],
+    })
+  )
+  .onCancel(() => {
```
cpreston321, line 181: I am wondering if it be good to add exitCode as the second param to get rid of the `process.exit(0)` or adding it within `p.cancel('Canceled', 0)`. What are your thoughts ?

Mist3rBru:
I agree, sounds more like an `workflow`, it could also change the `add` method to `step`.
Example:
```ts
await p.workflow()
  .step('name', () => p.text())
  .run()
```

cpreston321:
@Mist3rBru I like that. 

My only question here

1. Should we deprecate `groups` in favor on this ?  
2. Since this is the same functionality as `group` a end user might ask `Which one do I use ?`

Mist3rBru:
We can release it, and open a pinned issue like [v1.0.0 Roadmap](https://github.com/natemoo-re/clack/issues/35), so people can give their thoughts, and we decide based on it.

Mist3rBru:
I think it it a better API, since it does not deal with inference errors, and is more extensible.

packages/prompts/README.md
```diff
@@ -156,3 +156,66 @@ const group = await p.group(
...
+
+### Workflow
+
+Just like `group`, but on builder way, so you can choose which one fits better.
+
+```js
+import * as p from '@clack/prompts';
+
+const results = await p
+  .workflow()
```
cpreston321, line 168: @Mist3rBru You think it would be nice to have an option within workflow that `auto` ran the `run()` function to give user more flexibility?

cpreston321, line 168:
```suggestion
  .workflow({
  	auto: true
  })
```

or a single param.

```suggestion
  .workflow(true)
```

cpreston321:
@Mist3rBru I made some improvements to examples please check this out: https://github.com/natemoo-re/clack/commit/de9f5b6c7f92237189972f185b24d7128434f7d4

Mist3rBru:
@cpreston321 commit included.

This helps a lot to run the examples!

Issue #135, opened 2023-08-08
[Request] text: Add 'help' key to provide prompt description to the user


moniuch:
**Is your feature request related to a problem? Please describe.**
Could you suggest the best way to provide a legend/help text under the main prompt? I tried with new line, which probably would work for small things:

```
name: () => text({message: 'What is your name?\nDon\'t lie!'}),
```

but is not really a good solution for a couple of reasons.

![image](https://github.com/natemoo-re/clack/assets/681956/db27ac5e-4f32-4bc2-a372-8956a714a414)

I tried the `placeholder` option, but it is definitely meant for editable inputs, and not for what I'm after

![image](https://github.com/natemoo-re/clack/assets/681956/b4666c6f-fcf4-4d8c-98be-ebe32bffafee)

Would be best to have a hassle-free way of:
- aligning the hint with the main prompt
- wrapped text should also stay aligned with the main prompt
- dimmed style by default

The `placeholder` key is half-way there, I'd say - first line nicely aligned, dimmed out text.

![image](https://github.com/natemoo-re/clack/assets/681956/4629e4e3-5514-493b-8d39-1d317daa9a73)

It could be under a new key in the options object (`help`):

```
name: () => text({
  message: 'What is your name?',
  help: 'Lorem ipsum dolor sit amet, consectetur adipisicing elit. Atque eaque error eum ex, fugiat harum labore laboriosam libero magnam maiores modi natus optio quae quod recusandae repellendus sapiente vero voluptatum.'
}),
```

Motivation: in extensive flows for which the package might be used, explaining the step and the consequences of user decision, is essential, so that's why it was my first thing I tested here.

cpreston321:
Issue relates to: #111

Issue #111, opened 2023-03-21
[Request] Prompt description


joseph-farruggio:
**Is your feature request related to a problem? Please describe.**
Having the option to provide secondary text to provide more details or instruction would allow for more succinct prompts. 

**Describe the solution you'd like**
``` javascript
const myPrompt = await text({
    message: 'This is the primary question:',
    description: 'This is secondary text displayed between the message and where the user fills in the prompt`
})
```
**Describe alternatives you've considered**
I've tried writing longer prompts and getting clever with line breaks, but it takes away from Clack's gorgeous UI.

**Additional context**
N/A

Issue #35, opened 2023-02-14
v1.0.0 Roadmap


natemoo-re:
Infrastructure
- [ ] Testing (Vitest, needs to expose `input` and `output` to each prompt to mock)
- [ ] CI improvements
- [ ] More examples

Stability:
- [ ] Windows compat audit
- [ ] #20

Features:
- [ ] Additional Primitives
  - [ ] Grouped multiselect (Changesets-style)
  - [ ] Multiline text
  - [ ] Autocomplete / List search
  - [ ] File search?
- [ ] Prompt groups (like `prompts` / `inquirer`, but tree-shakeable)
- [ ] Improved Hooks #22

itpropro:
Any updates @natemoo-re ? Would love to use the autocomplete.
Is there anything you need help with to progress on v1?

natemoo-re:
In the interest of stability for existing `v0` users, [we've decided to switch `main` over to `v1.0.0-alpha.0`](https://github.com/bombshell-dev/clack/pull/250) and will be implementing new features on as part of the alpha release line.

Hoping to backport any critical bug fixes to the [`v0`](https://github.com/bombshell-dev/clack/tree/v0) releases, but no new minor releases.

dreyfus92:
@itpropro the autocomplete #288  has been merged.

ashleydavis:
Love this project and I'm keen to use it. Just wondering about the status of the `progress` component, will it be included in a  release soon?

43081j:
you should be able to install the 1.x alpha (`npm i @clack/prompts@alpha`), which has the progress prompt

its unlikely we will port those back to 0.x i think, but 1.x should actually be fairly stable if you want to try it 👍

ashleydavis:
> you should be able to install the 1.x alpha (`npm i @clack/prompts@alpha`), which has the progress prompt

Oh thanks so much! I wish had I noticed that much earlier.

> 
> its unlikely we will port those back to 0.x i think, but 1.x should actually be fairly stable if you want to try it 👍

I've already tried it. Given that I hadn't noticed the alpha version I pulled your entire codebase into my monorepo and used it directly ;)  Was just a test run anyway and I'll undo that, next step will be to install the alpha version and use that.

But yeah it's great so far, thanks for your work.

gstamp:
I think I might have hit a bug in the autocomplete. If I specify an `initialValue` option I end up getting this error: `TypeError: t.toLowerCase is not a function`

[...]

PR #141, opened 2023-08-12
feat(@clack/prompts): multiline support

### LINKED ISSUE #101: [Request] Multiline message

Currently, the left decorative line is not displayed when using `\n` line break.

Let message accept `string[]`, like this:

```ts
const msg = await confirm({
  message: ['Hello,', 'World!'],
}),
```

### LINKED ISSUE #132: [Bug] Spinner message is repeated infinitely if terminal window width is smaller than message width
**Environment**

- OS: MacOS
- Node Version: v16.14.0, also tested w/ Node 18 and Node 20
- Package: `@clack/prompts`
- Package Version: 0.6.3

**Describe the bug**
When the terminal in which a clack Spinner is run is smaller than the line length of the `spinner.start()` message, it is printed repeatedly to the terminal instead.  

**To Reproduce**
Minimal reproduction: https://stackblitz.com/edit/node-xdzhbp?file=index.js

- run `node index.js`
- resize terminal window with to be smaller than the spinner message length
==>  message is repeated infinitely

**Expected behavior**

The spinner message is not repeated infinitely but the original one is updated instead.

**Additional Information**

Our current workaround is to just print very short messages but ofc it'd be better to fix this properly.

Please let me know how we can help fixing this bug.

Mist3rBru:
# Multiline Support
Support multiline texts either with `\n`, or long ones that exceed terminal width, without breaking the layout

## List of supported prompts:
- [x] text
- [x] password
- [ ] confirm
    - Is just `yes`/`no`
- [x] select
- [ ] selectKey
    - It was made to handle short options
- [x] multiselect
- [ ] groupMultiselect
    - It was made to handle short options
- [x] note
- [x] intro
- [x] cancel
- [x] outro
- [x] log
- [x] spinner

## How to test
```
// From ~/clack
git remote add test-multiline https://github.com/Mist3rBru/clack.git
git checkout -b test-multiline
git pull test-multiline ml-example
pnpm build
cd examples/basic
npx jiti multiline.ts
```

Closes #101
Closes #132 
Relates #135 #111 #35

### DIFF
--- a/.changeset/popular-buses-buy.md
+++ b/.changeset/popular-buses-buy.md
@@ -1,0 +1,5 @@
+---
+'@clack/prompts': patch
+---
+
+Support multiline texts
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "organize-imports-cli": "^0.10.0",
     "prettier": "^2.8.4",
     "typescript": "^4.9.5",
-    "unbuild": "^1.1.2"
+    "unbuild": "1.1.2"
   },
   "packageManager": "pnpm@7.6.0",
   "volta": {
--- a/packages/prompts/src/index.ts
+++ b/packages/prompts/src/index.ts
@@ -75,7 +75,7 @@ function formatTextWithMaxWidth(
 	const endSymbol = options?.endSymbol ?? defaultSymbol;
 
 	const terminalWidth = process.stdout.columns || 80;
-	const maxWidth = options?.maxWidth ?? terminalWidth - 2;
+	const maxWidth = options?.maxWidth ?? terminalWidth;
 
 	const formattedLines: string[] = [];
 	const paragraphs = text.split(/\n/g);
@@ -85,8 +85,19 @@ function formatTextWithMaxWidth(
 		const words = paragraph.split(/\s/g);
 
 		for (const word of words) {
-			if (strLength(currentLine) + strLength(word) + 1 <= maxWidth) {
+			if (strLength(currentLine + word) + 3 <= maxWidth) {
 				currentLine += ` ${word}`;
+			} else if (strLength(word) + 2 >= maxWidth) {
+				const splitIndex = maxWidth - strLength(currentLine) - 3;
+				formattedLines.push(currentLine + ' ' + word.slice(0, splitIndex));
+
+				const chunkLength = maxWidth - 2;
+				let chunk = word.slice(splitIndex);
+				while (strLength(chunk) >= chunkLength) {
+					formattedLines.push(chunk.slice(0, chunkLength));
+					chunk = chunk.slice(chunkLength);
+				}
+				currentLine = chunk;
 			} else {
 				formattedLines.push(currentLine);
 				currentLine = word;
--- a/packages/prompts/src/index.ts
+++ b/packages/prompts/src/index.ts
@@ -60,13 +60,13 @@ const symbol = (state: State) => {
 
 function formatTextWithMaxWidth(
 	text: string,
-	options?: Partial<{
-		defaultSymbol: string;
-		initialSymbol: string;
-		newLineSymbol: string;
-		endSymbol: string;
-		lineWrapper: (line: string) => string;
-	}>
+	options?: {
+		defaultSymbol?: string;
+		initialSymbol?: string;
+		newLineSymbol?: string;
+		endSymbol?: string;
+		lineWrapper?: (line: string) => string;
+	}
 ): string {
 	const defaultSymbol = options?.defaultSymbol ?? color.cyan(S_BAR);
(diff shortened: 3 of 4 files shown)


cpreston321 (commented):
@Mist3rBru Straight killing it man 🙏🏼

Question: 

1. Shouldn't we support multiline at the core level vs trying make it work with `@clack/prompts`?

Mist3rBru:
I don't see how to implement it on `@clack/core`, since it strict resolves the issue when a new line doesn't start with its corresponding symbol from `@clack/prompts`

cpreston321:
I think we will need to pass a custom formatter func within the `render` function or expose a way for core to understand what is input vs styling. 

As I think this could fix support temporary for `prompts`,  I do feel like `core` will have to know so it can easily extended for other people not wanting to use `@clack/prompts`.

cpreston321:
Or we need to a way to expose what is the input text coming from core then do some special format around the input. I'm trying to think a neat way to handle it.

cpreston321:
@Mist3rBru I think for now you can keep this PR open and create a new draft pr that will mutate core functionality for a POW if you want. So we can get @natemoo-re to check it out. 

I'll do some more testing for the time being.

Mist3rBru:
What if we create a method on `core` Prompt called `format` or `style`, then all prompts can call `render() => this.format(msg)`

cpreston321:
Are you thinking of creating a new function within prompt called `format: (msg: string) => string` like `validator: ()`? If we go this route -- I think it would be good to make it super flexible. 

I am thinking of something that has a template then having a replacer for input styling. To keep things streamed line. (thinking out loud). Almost like template replacer with variables `{{title}}`, `{{input}}`. So it can be reused.

cpreston321:
@Mist3rBru Found some formatting issues and also a bug with the format function. It's a dim color for the | below the input and when selecting yes on install it errored out.

image

Mist3rBru:
@cpreston321 now must be right.

cpreston321:
Hey @Mist3rBru something still seems off: 

image
image

How it should look:
image
image
image

Mist3rBru:
@cpreston321 I forgot the cancel prompt

cpreston321:
Hey @Mist3rBru it looks like the default state looks off too it should look like this:

```bash
┌   create-app 
│
◇  Where should we create your project?
│  ./asdasd
│
```

not 

```bash
┌   create-app │
◆  Where should we create your project?
└  ./sparkling-solid
```

Mist3rBru:
@cpreston321 now must be right 🙏.

cpreston321:
@Mist3rBru I fixed the styles in this branch but if you want to copy them over then fix them here you can!

https://github.com/natemoo-re/clack/commit/e365d606fe46d7c8e05703a63eabffafa5ee704f

Mist3rBru:
Thanks @cpreston321, it is merged!

Now what should I do? Keep with this branch, merge it with the #143, or look for another approach to implement this feature.

cpreston321:
@Mist3rBru #143, 

I think it's a step in the right direction. I also think it might need some fine tuning for it to be fully there. 

In my head I would think there would be a template and spot for input to support the multi line support for formatting in the terminal almost like a wrapper and within a input. I just an easy way to leave styling up to the user and we can focus on core functionality. 

Within prompts we create a lot of duplicate styles but really the only thing that is the variant of the `input` and colors. If there was to define a base template/wrapper around the input, so that we can control the input with multi line support. 

I would think people would want to setup a custom wrapper for each prompt including state variants but also making the api very simple and easy to understand. Which I don't think is fully there. 

My thought (will fallback to base @clack/core `template/wrapper`) : 

```bash
 __________________________________
 |
 |  custom style template/wrapper   
 |
 | |-----------------------------|
 | |  input (core formating)     | 
 | |-----------------------------|
 |
 |
 |__________________________________ 
 ```
 
 I am just thinking out loud here but this is something I would imagine. I don't want the user to think about it too much.
 
 Making the api feel really easy, fluent, customizable is key.

cpreston321:
Also @Mist3rBru While playing around with multiline support I encounter this

https://github.com/natemoo-re/clack/assets/6619884/bf828677-2ea9-465b-b2ed-c14f579a3117

Have any idea ?

Mist3rBru:
It is almost fixed, there is an annoying `[7m` from cursor that shows only when the prev line is on last char that i coundn't solve
image

Mist3rBru:
@cpreston321 inner line break and cursor are fixed. I am going to close it, and focus on PR #143

cpreston321:
That sounds good, I think that will a better direction for the library -- since it will be at the core. I will add my input over there. 

Hopefully Nate can get some input over there too!

Thanks,
CP 🚀

PR #145, opened 2023-08-17
fix(@clack/core,@clack/prompts): add non empty opts type interface 

kevduc:
This PR addresses this issue https://github.com/natemoo-re/clack/issues/144.

The current implementation is erroring out at runtime, the implementation in this PR shows a type error in the IDE/tsc instead.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -8,3 +8,4 @@ export { default as SelectPrompt } from './prompts/select';
 export { default as SelectKeyPrompt } from './prompts/select-key';
 export { default as TextPrompt } from './prompts/text';
 export { block } from './utils';
+export type { NonEmptyArray } from './utility-types';
--- a/packages/core/src/prompts/group-multiselect.ts
+++ b/packages/core/src/prompts/group-multiselect.ts
@@ -1,14 +1,15 @@
+import { NonEmptyArray } from '../utility-types';
 import Prompt, { PromptOptions } from './prompt';
 
 interface GroupMultiSelectOptions<T extends { value: any }>
 	extends PromptOptions<GroupMultiSelectPrompt<T>> {
-	options: Record<string, T[]>;
+	options: Record<string, NonEmptyArray<T>>;
 	initialValues?: T['value'][];
 	required?: boolean;
 	cursorAt?: T['value'];
 }
 export default class GroupMultiSelectPrompt<T extends { value: any }> extends Prompt {
-	options: (T & { group: string | boolean })[];
+	options: NonEmptyArray<T & { group: string | boolean }>;
 	cursor: number = 0;
 
 	getGroupItems(group: string): T[] {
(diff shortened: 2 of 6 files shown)


kevduc:
Example of what the type error would look like:

Example

cpreston321:
Hey @kevduc ! Thanks for this PR. I recently merged a PR to simplify types on `@clack/prompts` if you want to resolve issues -- I can then be able to merge this! 

I could also see this being used also on https://github.com/natemoo-re/clack/blob/7c7fde8dcdfa1edc3feb131331290affbaa6d934/packages/prompts/src/index.ts#L300 & https://github.com/natemoo-re/clack/blob/7c7fde8dcdfa1edc3feb131331290affbaa6d934/packages/prompts/src/index.ts#L423.

kevduc:
@cpreston321 Thanks for the suggestions!  
The fix for `MultiSelectOptions` is pretty much the same as for select (https://github.com/natemoo-re/clack/pull/145/commits/20cde30e5e48fbe2b15a8d5dba2ae8f9845a4489).
I also had a look at `GroupMultiSelectPrompt`:
- it looks like it doesn't make sense to have an empty list of options for a given group (and it currenlty is buggy, see https://stackblitz.com/edit/node-51onk7?file=index.ts, the `test3` group looks selected, it can't be toggled, and when pressing enter it says nothing is selected), so the typing is fixed in https://github.com/natemoo-re/clack/pull/145/commits/f663d0f7e2a37ba7a088ac08f92c9e13a72dffe7
- it also looks like there's another bug when `options` itself is empty (see https://stackblitz.com/edit/node-dhmjec?file=index.ts, press space to make it crash):
  ```
  TypeError: Cannot read properties of undefined (reading 'group')
      at CD.toggleValue (file:///home/projects/node-dhmjec/node_modules/@clack/core/dist/index.mjs:40:1272)
      at Object.eval [as cb] (file:///home/projects/node-dhmjec/node_modules/@clack/core/dist/index.mjs:40:1049)
      at CD.emit (file:///home/projects/node-dhmjec/node_modules/@clack/core/dist/index.mjs:35:1851)
      at CD.onKeypress (file:///home/projects/node-dhmjec/node_modules/@clack/core/dist/index.mjs:35:2130)
  ```
  I tried an approach where I use a `NonEmptyObject` type on `options` in `GroupMultiSelectOptions`, i.e. something like this:
  ```ts
  // packages\core\src\utility-types.ts
  export type NonEmptyObject<T> = keyof T extends never ? never : T;
  
  // packages\core\src\prompts\group-multiselect.ts
  interface GroupMultiSelectOptions<
	  T extends { value: any },
	  U extends Record<string, NonEmptyArray<T>> = Record<string, NonEmptyArray<T>>
  > extends PromptOptions<GroupMultiSelectPrompt<T, U>> {
	  options: NonEmptyObject<U>;
	  initialValues?: T['value'][];
	  required?: boolean;
	  cursorAt?: T['value'];
  }
  ```
  but that pretty much breaks the inference of `Value` for `groupMultiselect` (it ends up always being `unknown`), so I've not added it to this MR, and if there's no straightforward fix, that's something that can be looked at in a different MR. An idea would be to at least do a runtime check and throw a more meaningful error.

43081j:
closing this as we're just going to enable strict index checks and handle empty arrays.

PR #151, opened 2023-08-26
feat(@clack/prompts): multiselect maxItems option

Mist3rBru:
This PR this introduces maxItems option to `multiselect` prompt, using the same logic as `select` prompt. Now both use the same utility function limitOptions that handles this logic.

### DIFF
--- a/.changeset/bright-rules-yell.md
+++ b/.changeset/bright-rules-yell.md
@@ -1,0 +1,5 @@
+---
+'@clack/prompts': patch
+---
+
+Feat multiselect maxItems option
--- a/packages/prompts/src/index.ts
+++ b/packages/prompts/src/index.ts
@@ -58,6 +58,41 @@ const symbol = (state: State) => {
 	}
 };
 
+interface LimitOptionsParams<TOption> {
+	options: TOption[];
+	maxItems: number | undefined;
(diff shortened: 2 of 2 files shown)


cpreston321:
Hey @Mist3rBru! Everything looks good besides the the submitted value. The old way would show something like this `@scope/a, @scope/b, @scope/c, @scope/z`.

vs 

image

Mist3rBru:
@cpreston321 This is from main branch, I did not changed this (state) behavior.

cpreston321:
Screenshot 1 was from main branch and screenshoot 2 was from your branch.

try it out in both: `pnpm dev`

Mist3rBru:
The screenshots are testing different cases, the [main:line62](https://github.com/natemoo-re/clack/blob/main/examples/changesets/index.ts#L62) is concatenating the result with `\n` on path step, I can change it, but I did't it.

cpreston321:
@Mist3rBru I thought it more of a symptom more then a test case. Since the maxItems was moved to a single function. 

So I figured it was it was a side effect. I don't think it was intentional on your side.

Mist3rBru:
@cpreston321 it is fine 😄. I gonna change this patch result to match multiselect result

Issue #152, opened 2023-08-26
[Request] Grouped spinners?


JoshuaKGoldberg:
**Is your feature request related to a problem? Please describe.**

Grouping (https://github.com/natemoo-re/clack/tree/24a3f4a20019c7dad2f69bcf163a8b5a3f747a16/packages/prompts#grouping) is available right now for prompts, but not for spinners. But I'd like to group the spinners together too: having a parent task and sub-task.

**Describe the solution you'd like**

Could we add a new spinner API analogous to `p.group`? Maybe, `p.spinnerGroup`?

```js
await p.spinnerGroup("Outer group", [
  ["First sub-task", async () => { /* ... */ }],
  ["Second sub-task", async () => { /* ... */ }],
  ["Third sub-task", async () => { /* ... */ }],
]);
```

**Describe alternatives you've considered**

In https://github.com/JoshuaKGoldberg/template-typescript-node-package/issues/708 I kind of did it manually, wrapping around Clack. It's not great.

https://github.com/natemoo-re/clack/assets/3335181/a3c6cbe4-8abe-4357-977a-377d4e53d9d1

**Additional context**

I'd be happy to contribute this, or if you don't want to keep it as a core feature, publish it as a standalone helper package.

JoshuaKGoldberg:
@cpreston321 that PR doesn't resolve this issue though. It just runs `s.start` and `s.stop` around the tasks. This issue is for _nesting_ tasks _within_ the spinners.

PR #154, opened 2023-08-27
feat(@clack/prompts): add tasks

artegoser:
Run tasks like listr.

This pr includes:
- [x] Code
- [x] Docs
- [x] Changeset

Maybe closes: #152

### DIFF
--- a/packages/prompts/src/index.ts
+++ b/packages/prompts/src/index.ts
@@ -781,11 +781,11 @@ export type Task = {
  */
 export const tasks = async (tasks: Task[]) => {
 	for (const task of tasks) {
-		if (task.enabled !== false) {
-			const s = spinner();
-			s.start(task.title);
-			const result = await task.task(s.message);
-			s.stop(result || task.title);
-		}
+		if (task.enabled === false) continue;
+
+		const s = spinner();
+		s.start(task.title);
+		const result = await task.task(s.message);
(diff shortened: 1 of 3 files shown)


packages/prompts/src/index.ts
```diff
@@ -759,3 +759,33 @@ export const group = async <T>(
...
+ * Define a group of tasks to be executed
+ */
+export const tasks = async (tasks: Task[]) => {
+	for (const task of tasks) {
+		if (task.enabled !== false) {
+			const s = spinner();
+			s.start(task.title);
+			const result = await task.task(s.message);
+			s.stop(result || task.title);
+		}
```
cpreston321, line 789:
```suggestion
		if (!task.enabled) continue;
		
		const s = spinner();
		s.start(task.title);
		const result = await task.task(s.message);
		s.stop(result || task.title);
```

A small suggestion as it's easier to read but Good work  @artegoser 🙏🏼

artegoser: Hello, but wouldn't you have to specify enabled property within each task? Maybe it would be better to do 
```ts
if (task.enabled === false) continue;
```

In this case, the task will be skipped only if the user explicitly specified false.

cpreston321 (approved):

lianghx-319:
Please release this features

PR #162, opened 2023-09-08
chore(@clack/prompts): update readme

await-ovo:
Currently if we just copy the example code of [cancellation](https://github.com/natemoo-re/clack/tree/main/packages/prompts#cancellation), it will throw an error:

```
TypeError: Cannot read properties of undefined (reading 'validate')
```

It might be useful to fill in the required parameters here to ensure that newbies don't get an error when copying this code~


natemoo-re (approved):
Sorry for the delay, thanks for the PR!

PR #169, opened 2023-09-22
feat(@clack/prompts): adapt `spinner` to CI environment

### LINKED ISSUE #168: [Request] Disable continuous terminal rewriting in CI

Spinning out from https://github.com/JoshuaKGoldberg/create-typescript-app/issues/859: when running Clack's spinners in a project in CI (e.g. GitHub Actions), you get a _ton_ of CI logs:

```plaintext
│
◇  ✅ Passed migrating repository structure.
[25](https://github.com/JoshuaKGoldberg/create-typescript-app/actions/runs/6264408991/job/17011040698?pr=858#step:5:26)l│
◒  Cleaning up files◐  Cleaning up files◓  Cleaning up files◑  Cleaning up files◒  Cleaning up files◐  Cleaning up files◓  Cleaning up files◑  Cleaning up files◒  Cleaning up files.◐  Cleaning up files.◓  Cleaning up files.◑  Cleaning up files.◒  Cleaning up files.◐  Cleaning up files.◓  Cleaning up files.◑  Cleaning up files.◒  Cleaning up files..◐  Cleaning up files..◓  Cleaning up files..◑  Cleaning up files..◒  Cleaning up files..◐  Cleaning up files..◓  Cleaning up files..◑  Cleaning up files..◒  Cleaning up files...◐  Cleaning up files...◓  Cleaning up files...◑  Cleaning up files...◒  Cleaning up files...◐  Cleaning up files...◓  Cleaning up files...◑  Cleaning up files...◒  Cleaning up files...◐  Cleaning up files◓  Cleaning up files◑  Cleaning up files◒  Cleaning up files◐  Cleaning up files◓  Cleaning up files◑  Cleaning up files◒  Cleaning up files◐  Cleaning up files.◓  Cleaning up files.◑  Cleaning up files.◒  Cleaning up files.◐  Cleaning up files.◓  Cleaning up files.◑  Cleaning up files.◒  Cleaning up files.◐  Cleaning up files..◓  Cleaning up files..◑  Cleaning up files..◒  Cleaning up files..◐  Cleaning up files..◓  Cleaning up files..◑  Cleaning up files..◒  Cleaning up files..◐  Cleaning up files...◓  Cleaning up files...◑  Cleaning up files...◒  Cleaning up files...◐  Cleaning up files...◓  Cleaning up files...◑  Cleaning up files...◒  Cleaning up files...◐  Cleaning up files...◓  Cleaning up files◑  Cleaning up files◒  Cleaning up files◐  Cleaning up files◓  Cleaning up files◑  Cleaning up files◒  Cleaning up files◐  Cleaning up files◓  Cleaning up files.◑  Cleaning up files.◒  Cleaning up files.◐  Cleaning up files.◓  Cleaning up files.◑  Cleaning up files.◒  Cleaning up files.◐  Cleaning up files.◓  Cleaning up files..◑  Cleaning up files..◒  Cleaning up files..◐  Cleaning up files..◓  Cleaning up files..◑  Cleaning up files..◒  Cleaning up files..◐  Cleaning up files..◓  Cleaning up files...◑  Cleaning up files...◒  Cleaning up files...◐  Cleaning up files...◓
```

Screenshot of the preceding code block in CI, showing many "Cleaning up files.." notices

Could we an option to disable terminal animations, so each line is only logged once? Ideally that option would default to auto-detecting whether the environment is a CI one.

User-land code could avoid spinners in CI by doing the detection themselves. That's a lot of work for each consuming project though.

**Additional context**

I'd be happy to implement this - though might not have time soon.

orochaa:
This PR addresses a issue reported in GitHub Actions where `spinner` was excessively writing messages, leading to confusion and cluttered output. To enhance the CI workflow and provide a smoother experience, the following changes have been made only for CI environment:

- Messages will now only be written when a `spinner` method is called and the message updated, preventing unnecessary message repetition.
- There will be no loading dots animation, instead it will be always `...`
- Instead of erase the previous message, action that is blocked during CI, it will just write a new one.

### Demo
https://github.com/natemoo-re/clack/assets/100330057/0997f28f-0677-4223-b81a-db47776cc858

Closes #168


ulken:
Suggestion: use something like https://www.npmjs.com/package/ci-info to cover more CI environments.

examples/basic/package.json
```diff
@@ -10,7 +10,8 @@
   },
   "scripts": {
     "start": "jiti ./index.ts",
-    "spinner": "jiti ./spinner.ts"
+    "spinner": "jiti ./spinner.ts",
+    "spinner-ci": "npx cross-env GITHUB_ACTIONS=\"true\" jiti ./spinner-ci.ts"
```
antony, line 14:
```suggestion
    "spinner-ci": "npx cross-env CI=\"true\" jiti ./spinner-ci.ts"
```

packages/prompts/package.json
```diff
@@ -55,7 +55,8 @@
   "dependencies": {
     "@clack/core": "workspace:*",
     "picocolors": "^1.0.0",
-    "sisteransi": "^1.0.5"
+    "sisteransi": "^1.0.5",
+    "std-env": "^3.4.3"
```
antony, line 59: this is an entire new package simply to check an environment variable. I've yet to find a CI platform which doesn't set an env var called CI=true.

If this library hopes to retain it's small size, it probably makes sense not to add this dependency, as I don't think it is required.

antony (dismissed):
Suggest that the new std-env dependency is removed as it probably isn't required. Generally, CI systems set CI=true, and I've yet to encounter one which doesn't.

If some random CI system doesn't set CI=true, then it's trivial for a user to just set it.

But in addition, if somebody wanted the old behaviour in CI, they could override it using this variable - whereas the `std-env` library checks a whole pile of other variables, making that hard to do.

In addition,

Mist3rBru:
Hey @antony, thanks for the review!

I agree with the suggestions and have applied then to the PR, now we just check if CI variable is set to true. This way we cover most CI environments and still keep the bundle size small.

.changeset/thin-moose-tease.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clack/prompts': patch
+---
+
+Adapt `spinner` to CI environment
```
natemoo-re, line 5: Adapts `spinner` output for static CI environments

natemoo-re (approved):
Thanks for the PR! This is great 💜

mdjastrzebski:
@natemoo-re when do you plan to release this fix?

PR #180, opened 2023-11-28
chore: improve tsconfig

orochaa:
This pull request includes updates to the tsconfig.json file, aligning it with modern development practices. The changes introduced are as follows:

## Changes
- `"moduleResolution": "Bundler"`: This option is configured for use with bundlers. Unlike the Node.js resolution modes, this setting supports package.json "imports" and "exports" without requiring file extensions on relative paths in imports.
- `"moduleDetection": "force"`: Enabling this option ensures that every non-declaration file is treated as a module, providing a more consistent and predictable module handling approach.
- `"isolatedModules": true`: TypeScript will now issue warnings for code that might not be correctly interpreted by a single-file transpilation process, helping catch potential issues early in the development process.
- `"verbatimModuleSyntax": true`: Forces type imports to be flagged with a `type` modifier. This is particularly useful for bundlers, as it signals upfront that the code won't be used as a value, allowing the bundler to exclude it from the generated JavaScript code.
- `"lib": ["ES2022"]`: Updated to include only APIs from ES2022, explicitly excluding DOM APIs.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -3,11 +3,15 @@
     "noEmit": true,
     "module": "ESNext",
     "target": "ESNext",
-    "moduleResolution": "node",
+    "moduleResolution": "Bundler",
+    "moduleDetection": "force",
     "strict": true,
     "esModuleInterop": true,
     "forceConsistentCasingInFileNames": true,
     "skipLibCheck": true,
+    "isolatedModules": true,
+    "verbatimModuleSyntax": true,
+    "lib": ["ES2022"],
     "paths": {
       "@clack/core": ["./packages/core/src"]
     }


tsconfig.json
```diff
@@ -3,11 +3,15 @@
     "noEmit": true,
     "module": "ESNext",
     "target": "ESNext",
-    "moduleResolution": "node",
+    "moduleResolution": "Bundler",
```
benmccann, line 6: `bundler` is good for apps, but `node16` is better for libraries. that's because bundler is more permissive about things like file endings, which is good for consuming a variety of libraries, but `node16` is stricter, which is better when your library may be consumed by apps with a variety of different settings

Mist3rBru: Yes, but in the end it will be changed by `unbuild`. Do you think it's still worth it?

benmccann: I'm not familiar with `unbuild` or why they would do that. Ideally it shouldn't or there should be an option to configure it

Mist3rBru: `unbuild` imports all the code and puts it in a single file, then minimizes it, so there is no import after it has been compiled. It means that the `moduleResolution` won't have any impact after all.

Mist3rBru:
> This pull request includes updates to the tsconfig.json file, aligning it with modern development practices. The changes introduced are as follows:
> 
> ## Changes
> * `"moduleResolution": "Node16"`: It was updated from `node` option, which was deprecated.
> * `"moduleDetection": "force"`: Enabling this option ensures that every non-declaration file is treated as a module, providing a more consistent and predictable module handling approach.
> * `"isolatedModules": true`: TypeScript will now issue warnings for code that might not be correctly interpreted by a single-file transpilation process, helping catch potential issues early in the development process.
> * `"verbatimModuleSyntax": true`: Forces type imports to be flagged with a `type` modifier. This is particularly useful for bundlers, as it signals upfront that the code won't be used as a value, allowing the bundler to exclude it from the generated JavaScript code.
> * `"lib": ["ES2022"]`: Updated to include only APIs from ES2022, explicitly excluding DOM APIs.

natemoo-re (approved):
We might end up refactoring away from `unbuild` and have to change this, but it seems fine for now! No strong opinion on these ones.

PR #182, opened 2023-12-14
fix(@clack/core): keyboard input not working after await in spinner

### LINKED ISSUE #176: [Bug]
**Environment**

- OS: Windows
- Node Version: v18.17.1
- Package: @clack/prompts
- Package Version: v0.7.0]

**Describe the bug**
After updating to the latest version, prompt behavior has changed after awaiting spinner. The await line between spinner start and stop is causing issues with the second following prompt, where it does not respond to any key input (like arrow keys or even Ctrl-C), only the [Enter] key, which then "activates" the normal behavior, but creates a duplicate of the first option.
With the await line commented out, everything works as expected.
The prompt order does not matter, the second one is always effected.
```
│
◇  Test anyway?
│  Yes
│
◆  test
│  ◻ A (hint A)
│  ◻ A
│  ◻ B (hint B)
│  ◻ C
└
```
```
◇  test
│  B
│
◆  Test anyway?
│  ● Yes / ○ No
│  ○ Yes / ● No
```

**To Reproduce**

- Using the following code:
```js
  const sp = p.spinner();
  sp.start("start");
  await setTimeout(2000);
  sp.stop("stop");

  const y = await p.confirm({
    message: "Test anyway?",
  });

  const x = await p.multiselect({
    message: "test",
    options: [
      { value: "a", label: "A", hint: "hint A" },
      { value: "b", label: "B", hint: "hint B" },
      { value: "c", label: "C", hint: "hint C" },
    ],
    required: false,
  });
```

**Expected behavior**
To be able to navigate from prompt to prompt using the keyboard, and use Ctrl-C to quit.

Mist3rBru:
Fix #176 

## Description
The issue occurred on Windows, when the await line between the spinner start and stop was present, causing the second following prompt to be unresponsive to key inputs (e.g., arrow keys or Ctrl-C), except for the [Enter] key. Additionally, this resulted in the duplication of the first option.

## Solution 
The solution is identify when it is running on Windows and do not change terminal's mode
```ts
if (input.isTTY && !isWindows) input.setRawMode(false);
```

### Tested Environment:

OS: Windows, Ubuntu
Node Version: v18.17.1
Package: @clack/prompts
Package Version: v0.7.0

## Snippet to Reproduce
```ts
const sp = p.spinner();
sp.start("start");
await setTimeout(2000);
sp.stop("stop");

const y = await p.confirm({
  message: "Test anyway?",
});

const x = await p.multiselect({
  message: "test",
  options: [
    { value: "a", label: "A", hint: "hint A" },
    { value: "b", label: "B", hint: "hint B" },
    { value: "c", label: "C", hint: "hint C" },
  ],
  required: false,
});
```

### DIFF
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -4,6 +4,8 @@ import { stdin, stdout } from 'node:process';
 import * as readline from 'node:readline';
 import { cursor } from 'sisteransi';
 
+const isWindows = globalThis.process.platform.startsWith('win');
+
 export function block({
 	input = stdin,
 	output = stdout,
@@ -40,7 +42,9 @@ export function block({
 	return () => {
 		input.off('keypress', clear);
 		if (hideCursor) process.stdout.write(cursor.show);
-		if (input.isTTY) input.setRawMode(false);
+
+		// Prevent Windows specific issues: https://github.com/natemoo-re/clack/issues/176
+		if (input.isTTY && !isWindows) input.setRawMode(false);
 
 		// @ts-expect-error fix for https://github.com/nodejs/node/issues/31762#issuecomment-1441223907
 		rl.terminal = false;
--- a/.changeset/olive-walls-clean.md
+++ b/.changeset/olive-walls-clean.md
@@ -1,0 +1,5 @@
+---
+'@clack/core': patch
+---
+
+fix(@clack/core): keyboard input not working after await in spinner


lucas-zimerman:
that indeed fixes the issues, I just wonder why it only affects the second

lucas-zimerman:
We actually don't need `input.setRawMode`
The thing is that clear here is being used for two flows  (exiting the program if ctrl-c is pressed and clearing the line) instead of one (clearing  the line).

On linux, `SIGINT` should be triggered so we can hook to it (I'll have to check it)

I'll add my suggestions on separated lines so that it can be marked as solved if implemented or discuss specifics parts of the suggestion.

packages/core/src/utils.ts
```diff
@@ -40,7 +42,9 @@ export function block({
 	return () => {
 		input.off('keypress', clear);
 		if (hideCursor) process.stdout.write(cursor.show);
-		if (input.isTTY) input.setRawMode(false);
```
lucas-zimerman, line 43: I would recommend removing both references to `input.setRawMode` since it it's not required for clearing the screen and checking for sigint.

packages/core/src/utils.ts
```diff
@@ -40,7 +42,9 @@ export function block({
 	return () => {
 		input.off('keypress', clear);
```
lucas-zimerman, line 43: I would suggest to rename the const `clear` to `sigintOrClear` so it reads that this functions does more than just clearing the screen

Lms24:
Hey @natemoo-re friendly bump: any chance this PR could get reviewed and merged in?

natemoo-re (approved):
Thank you for the fix and thanks for your patience!

rorry121:
Any news？When will a new version be released? @natemoo-re

PR #195, opened 2024-06-04
test(core): add initial tests

43081j:
Adds some tests for the core package and fixes a small bug found during writing them.

The bug is basically that we never unhide the cursor on `process.exit` conditions, meaning we leave it hidden in the terminal after our prompt has been cancelled.

cc @natemoo-re  i was trying to get my head around how this codebase works etc so wrote these while i was in there. they're not as nice as some kind of e2e cli test i suppose, but still seem straight forward enough (and better than nothing)

### DIFF
--- a/.changeset/gold-dryers-bake.md
+++ b/.changeset/gold-dryers-bake.md
@@ -1,0 +1,5 @@
+---
+"@clack/core": patch
+---
+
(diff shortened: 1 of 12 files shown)


natemoo-re (approved):
Sorry I'm just getting to this! Absolutely lovely and very helpful, thanks for scaffolding these out! 🙌

Issue #36, opened 2023-02-14
[feature]: @clack/prompts support for many different themes, variants, and colors


cpreston321:
**Is your feature request related to a problem? Please describe.**

No

**Describe the solution you'd like**

Since prompts is super nice it would be since to see a way to either change the `theme` of the prompts either containing theme within another repo that has those theme, but since prompts is on color and preset of `icons` you could create a way to customize the colors through config or maybe some type of system that can define these as a variant or theme. Since `prompts ` is more a granular term but could be perceived as an all in one alternative solution for CLI prompts.

**Describe alternatives you've considered**

N/A

**Additional context**

I would like to discuss more here but it's an idea. Even though `@clack/prompts` is a wrapper for `core` I foresee it being bigger package that is built off of core that can include many handy utils, functions, and prompts. Maybe it's the wrong package for it. I am kinda thinking of variants like `astro-theme`, `clack-theme` and these will drive icons, styles etc.

natemoo-re:
I like this idea a lot and have similar thoughts. I haven't really sketched it out yet, but totally agree that theming would be a really powerful feature.

If you have any API ideas I'm all ears!

cpreston321:
I think theming can go many ways like if you want the theme layer to be at the `core` package and have a class that can drive themes. The could potentially be the base point that then, you can extend it within prompts that will create a bigger theme for prompts and then add custom variants/themes to inputs that was wrapped within prompts.

The only problem with injecting a theme within core is that you will have multiple prompts with the same config. So goes back to whats the best way to apply global settings that will look at the applied settings. You could possibly use `unjs/c12` smart configuration loader from the root of the directory. To then load a config file of your choice to then drive global configuration. Or it could be at the core level but since this is a CLI root doesn't seem bad to have something at the root level config that can drive settings.

kevinwolfcr:
@natemoo-re I think you could benefit from the NodeJS's [`AsyncLocalStorage`](https://nodejs.org/api/async_hooks.html#class-asynclocalstorage) class, and use something like React Context to propagate the user-defined theme through all the prompts. So the API could be something like this:

```ts
import { run } from `@clack/prompts`

run(async () => {
  // your cli code here...
}, { theme: someThemeConfig, onCancel: someCustomCancelHandling })
```

pi0:
[...]

PR #212, opened 2024-12-02
feat(@clack/prompts): configurable colors

hazycora:
A partial implementation of #36.

I'm sure the maintainers likely have their own ideas for how this could work instead. I've made this _for my own use_ until #36 is resolved, but just on the off-chance my approach is okay with the maintainers here I thought I'd try submitting it as a PR. Fully understand if this approach isn't preferred :) It's highly customisable to an extent that could be seen as unnecessary, lol.

Colors for specific contexts are defined in a `config.colors` exported object, which users of the library can edit. If, for instance, you wanted to replace all uses of green in the default theme with magenta, you could do this:

```ts
import * as p from '@clack/prompts'

p.config.colors.symbol.submit = 'magenta'
p.config.colors.log.success = 'magenta'
p.config.colors.log.step = 'magenta'
p.config.colors.radio.active = 'magenta'
p.config.colors.checkbox.selected = 'magenta'
p.config.colors.key.active = ['inverse', 'bgMagenta', 'gray']
```

I've used a config object instead of just having an object for the colors themselves, in case more was configurable in the future, like the symbols used for various things.

### DIFF
--- a/packages/prompts/src/index.ts
+++ b/packages/prompts/src/index.ts
@@ -8,10 +8,11 @@ import {
 	SelectKeyPrompt,
 	SelectPrompt,
 	State,
-	TextPrompt
+	TextPrompt,
 } from '@clack/core';
 import isUnicodeSupported from 'is-unicode-supported';
 import color from 'picocolors';
+import { Colors } from 'picocolors/types';
 import { cursor, erase } from 'sisteransi';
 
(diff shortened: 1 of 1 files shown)


43081j:
we're going to solve this another way by having a `theme`, so im going to close this

thank you for the help still 🙏

PR #230, opened 2025-01-27
feat: timer indicator for `spinner`

mdjastrzebski:
## Summary

Adds spinner option for showing timer instead of animated dots at the end `...` .

```
spinner() // defaults to dots for compatibility
spinner({ indicator: `dots` }) // explicitly set dots
spinner({ indicator: `timer` }) // enable timer indicator
```

Manually adding timer (or any other custom progress indicator) results in too verbose logging on CI, as each partial update results in new spinner message.

## Visualization
https://github.com/user-attachments/assets/e7e3b52c-ec55-444c-9e88-e3a64fe07417

## Testing

Run `pnpm spinner-timer` in `examples/basic`. Also check `spinner` and `spinner-ci` to ensure that default behavior is not broken.

## Discussion points

1. This could alternatively be implemented as a more general progress indicator where user would provide the indicator callback to run and which would be ignored by the CI. 
2. Should `...` loading dots be kept when timer is running? Currently they have a different cadence, which creates a weird effect. If we were to keep them it would make sense to sync them to the timer, i.e. one do per second (or maybe 1/2 second)


afrianjunior:
Thats good idea!

natemoo-re (approved):
Great job, this is a very nice addition! Releasing in the next minor.

.changeset/bright-chefs-double.md
```diff
@@ -0,0 +1,5 @@
+---
+'@clack/prompts': minor
+---
+
+spinner timer progress indicator
```
natemoo-re, line 5:
```suggestion
Adds a new `indicator` option to `spinner`, which supports the original `"dots"` loading animation or a new `"timer"` loading animation.

```ts
import * as p from '@clack/prompts';

const spin = p.spinner({ indicator: 'timer' });
spin.start('Loading');
await sleep(3000);
spin.stop('Loaded');
```

PR #233, opened 2025-02-16
fix: remove types condition

bluwy:
The types condition can't be used for both ESM and CJS entrypoints, they need to be separated.

- https://publint.dev/@clack/core@0.4.1
- https://arethetypeswrong.github.io/?p=%40clack%2Fcore%400.4.1

Removing the types condition can fix this as it'll resolve to `./dist/index.mjs` or `./dist/index.cjs`, and then TypeScript can infer the sibiling `./dist/index.d.mts` and `./dist/index.d.cts` files automatically (both already published). Explicitly specifying two types condition is also possible but this change felt simpler.

### DIFF
--- a/.changeset/breezy-trees-fetch.md
+++ b/.changeset/breezy-trees-fetch.md
@@ -1,0 +1,6 @@
+---
+"@clack/prompts": patch
+"@clack/core": patch
+---
+
+Fix types export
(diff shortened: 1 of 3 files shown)


dreyfus92:
hey @bluwy, thanks for the PR! just a heads-up — this was recently addressed when we moved to ESM-only in [#246](https://github.com/bombshell-dev/clack/pull/246), which resolves the issue by removing CJS altogether, so I think we're good on this front now. appreciate you taking the time to fix it though!

bluwy:
I believe the types still need to be fixed in main. If .mjs is used, the the types should be .d.mts, not .d.ts.

Issue #192, opened 2024-04-29
[Bug] ERR_TTY_INIT_FAILED on git bash (MINGW64)


brc-dd:
**Environment**

- OS: Windows
- Node Version: v22.0.0
- Package: `@clack/prompts`
- Package Version: v0.7.0

**Describe the bug**

https://github.com/natemoo-re/clack/blob/45ee73bf33b25f9a8c7e1bb6117ccc165478bf4d/packages/core/src/prompts/prompt.ts#L82

This line throws `ERR_TTY_INIT_FAILED` when used inside terminals like Git Bash.

**To Reproduce**

You'll need a Windows machine (or VM, I'm using UTM on mac) with Git Bash installed. Then try running any project using clack. For example with VitePress:

```sh
mkdir foo
npm add -D vitepress
npx vitepress init
``` 

Or with Svelte:

```sh
npm create svelte@latest foo
```

**Expected behavior**

Work similar to other terminals or maybe a fallback strategy in case write stream can't be opened on stdin.

**Additional Information**

Downstream issues - https://github.com/vuejs/vitepress/issues/3841, https://github.com/sveltejs/svelte/issues/8595, https://stackoverflow.com/q/75750730/11613622,

Dmytro-Shulha:
Having same problem when trying to run tests from within Git Bash:
![image](https://github.com/user-attachments/assets/2a9ee324-88ea-42de-bd81-1d9c94de78df)

natemoo-re:
Hey, thanks for opening an issue! I'm finally working through the backlog here.

Windows strikes again... I'll have to investigate what other libraries do to workaround this! All of Windows support is honestly pretty naive and it'd be great to do more feature detection in general.

lilnasy:
The root cause is the fact that the version of `mintty` terminal emulator included with git on windows is not interactive by default.

![Image](https://github.com/user-attachments/assets/50e7e7bf-c9d2-4ee9-8a4d-a12ff7c283c9)

Node.js contributor documentation makes a [footnote](https://github.com/nodejs/node/blob/338d70b752fd918ebd568f432373065e04539830/BUILDING.md#user-content-fn-2-9f1c6f9f869ee1d34ac9b68befcfc525) of this:
[...]

PR #235, opened 2025-02-18
fix: uv_tty_init error on Windows

sapphi-red:
This PR fixes the following error happening on Windows for non-tty terminals.

> TTY initialization failed: uv_tty_init returned EBADF (bad file descriptor)

The error was happening this line.
https://github.com/bombshell-dev/clack/blob/d6d9ce791e33dbd40191a47f12db74d19bfbdb8d/packages/core/src/prompts/prompt.ts#L136
It seems calling `new tty.WriteStream` with non-tty file descriptor throws the error above.

I changed `tty.WriteStream` with `stream.Writable` and set `terminal: true` option for `readline.createInterface` so that readline will treat `stream.Writable` as a terminal.

I first went with replacing `new WriteStream(0)` with `isatty(0) ? new WriteStream(0) : fs.createWriteStream(null, { fd: 0 })`. It worked locally, but that didn't pass the tests.

refs #192

### DIFF
--- a/.changeset/lemon-monkeys-help.md
+++ b/.changeset/lemon-monkeys-help.md
@@ -1,0 +1,5 @@
+---
+"@clack/core": patch
+---
+
+Fix "TTY initialization failed: uv_tty_init returned EBADF (bad file descriptor)" error happening on Windows for non-tty terminals.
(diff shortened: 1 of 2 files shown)


Zamiell:
@sapphi-red Thanks so much! I am trying out clack for the first time on Windows, using the built-in Git Bash for Windows that comes automatically when you install Git, and I was confronted with this error.

Here's hoping we can get this merged soon with a new release!

Issue #25, opened 2023-02-14
Feature request : Multiline text


lgersman:
Your project looks promising !

Would you consider to add a Multiline Text Input like in Gum (https://github.com/charmbracelet/gum#write) ?

natemoo-re:
Yep, I would love to add this! Looking at our current logic, most things are expecting a single line value, so this might take some internal refactoring.

SebghatYusuf:
any updates here?

deadcoder0904:
Would love this too bdw.

natemoo-re:
🙈 better 1,170 days late than never—the multiline prompt has been released in [`@clack/prompts@1.3.0`](https://github.com/bombshell-dev/clack/releases/tag/%40clack%2Fprompts%401.3.0)

PR #240, opened 2025-02-27
feat(@clack/core,@clack/prompts): Multiline text input

MacFJA:
Add a new prompt: `MultiLinePrompt` and its rendering `multiline`

### Added to `@clack/core`

- Option to change the key that submit a prompt (default value is `return` to keep retro-compatibility)
- A new event that send the raw char and key
- A new util function (`isSameKey`) to test if two key are the same (with `ctrl`, `meta`, `shift` included in the check)
- A new prompt `MultiLinePrompt`

### Added to `@clack/prompts`

- A new prompt function `multiline`

## Difficulties and workaround

**Issue**: The `return` key is used to submit the prompt and is hardcoded, so impossible to insert new line
**Solution**: Make the submit key configurable (see: `packages/core/src/prompts/prompt.ts:{22,53,210}`) 

**Issue**: The default cursor management use vim mapping, so when pressing letter `k`, `j`, `h` or `l` to add those letter, the cursor also moves.
**Solution**: Re-implementing of the cursor management without vim mapping (see `packages/core/src/prompts/multi-line.ts:29`)

**Issue**: The event `key` lower the inputted char, so uppercased char are never displayed
**Solution**: Create a new event `rawKey` to send the untouched value (also used for the cursor management)

---

- This PR (except if I miss something) don't create incompatibility with existing usage of the library (as it add new feature, and core changes have default value that behave as it was before)
- This PR fix the issue #25


natemoo-re (approved):
Wow great stuff, thank you for contributing!

deadcoder0904:
I'd love if this is merged. Its annoying to do multi-line with such a beautiful library. Only thing missing imo.

dreyfus92:
thank you @43081j 🎉

PR #241, opened 2025-02-28
feat(@clack/prompts): add help on prompt

### LINKED ISSUE #135: [Request] text: Add 'help' key to provide prompt description to the user

MacFJA:
Add help/description option on:
- text
- password
- confirm
- select
- multiselect
- groupMultiselect

<details open>
<summary>Screenshots</summary>

Image
Image
Image

</details>

---

- The help text is display in dim and italic to differentiate from the placeholder
- The help text doesn't appear on submitted prompt or canceled prompt

---

I'm not sure about the icon
Right now, the icon is [ⓘ](https://symbl.cc/en/24D8/)

Other options were:
image

---

Fix #135 and #111

### DIFF
--- a/packages/prompts/src/index.ts
+++ b/packages/prompts/src/index.ts
@@ -27,6 +27,7 @@ const S_STEP_SUBMIT = s('◇', 'o');
 
 const S_BAR_START = s('┌', 'T');
 const S_BAR = s('│', '|');
+const S_BAR_DETAIL = s('│ⓘ', `|${color.italic('i')}`);
 const S_BAR_END = s('└', '—');
 
 const S_RADIO_ACTIVE = s('●', '>');
@@ -58,6 +59,11 @@ const symbol = (state: State) => {
 		case 'submit':
 			return color.green(S_STEP_SUBMIT);
 	}
+};
+
+const help = (text: string | undefined, style: (input: string) => string) => {
+	if (text === undefined) return '';
+	return `${style(S_BAR_DETAIL)} ${color.dim(color.italic(text))}\n`;
 };
 
 interface LimitOptionsParams<TOption> {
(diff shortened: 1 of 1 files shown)


packages/prompts/src/index.ts
```diff
@@ -27,6 +27,7 @@ const S_STEP_SUBMIT = s('◇', 'o');
 
 const S_BAR_START = s('┌', 'T');
 const S_BAR = s('│', '|');
+const S_BAR_DETAIL = s('│ⓘ', `|${color.italic('i')}`);
```
natemoo-re, line 30: Flagging this since I want to play with it locally and come up with a good indicator. Having an icon on the inside margin feels a bit off, but I don't know what will feel better just yet.

Thanks for including your explorations in the PR description!

natemoo-re (approved):
Thanks for implementing this! Looks great. I'm going to take another pass at the icon before merging.

MacFJA:
I did some other test that don't use an icon at all:
image

dreyfus92 (commented):
@MacFJA this has become stale, do you mind revisiting this?

PR #242, opened 2025-02-28
fix(@clack/prompts): handle spinner.stop call when spinner.start was not called

Minishlink:
Basic reproduction would be

```ts
const spin = spinner();
try {
  await doSomethingWithAnotherSpinnerInsideAndThrows();
  spin.start('loading do');
  await doSomething()
  spin.stop('success!')
} catch(error) {
  spin.stop('error!', 2)
  handleError(error)
}
```

### DIFF
--- a/packages/prompts/src/index.ts
+++ b/packages/prompts/src/index.ts
@@ -729,7 +729,7 @@ export const spinner = ({ indicator = 'dots' }: SpinnerOptions = {}) => {
 	const delay = unicode ? 80 : 120;
 	const isCI = process.env.CI === 'true';
 
-	let unblock: () => void;
+	let unblock: (() => void) | undefined;
 	let loop: NodeJS.Timeout;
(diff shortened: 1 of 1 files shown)


DogPawHat:
Wouldn't it be better if it threw an error instead if `stop` was called before `start`?

```
if (!isSpinnerActive) {
	throw new Error('You can not stop the spinner if it is not active');
}
```

Minishlink:
I'd rather make .stop return early if !isSpinnerActive but I'm ok with this if you can access isSpinnerActive publicly

43081j:
i think we should return early like you say

so keep `unblock` as it is (because if we reach it, it must've been initialised), but change `stop` to return early:

```ts
if (!isSpinnerActive) {
  return;
}
```

43081j:
looks like this was fixed in 0b852e10

PR #245, opened 2025-03-07
fix: weaken internal group types to strengthen return/onCancel types

43081j:
This changes a few things.

First of all, trying to infer `T` while defining `T` currently doesn't work unless there's a member without the `results` arg. This is a known limitation that may or may not be fixed in typescript one day.

Until that happens, if ever, this dumbs down the type of `results` to be a `Record<PropertyKey, unknown>`. This means _within a function_ you lose strong typing, but the return type will now always be correct.

Secondly, `T` is actually the resulting already-awaited shape. This means we do not need `Awaited<T[K]>` since `T[K]` is already the awaited type.

For example:

```ts
type ActualType = {
  foo: number;
  bar: number;
};
group<ActualType>({
  foo: () => Promise.resolve(303),
  bar: () => Promise.resolve(808)
});
```

You can see the `ActualType` never needed `Awaited<T[K]>` on each type since it is already the final result.

### DIFF
--- a/packages/prompts/src/index.ts
+++ b/packages/prompts/src/index.ts
@@ -844,16 +844,12 @@ export const spinner = ({ indicator = 'dots' }: SpinnerOptions = {}) => {
 	};
 };
 
-export type PromptGroupAwaitedReturn<T> = {
-	[P in keyof T]: Exclude<Awaited<T[P]>, symbol>;
-};
-
 export interface PromptGroupOptions<T> {
 	/**
 	 * Control how the group can be canceled
 	 * if one of the prompts is canceled.
 	 */
-	onCancel?: (opts: { results: Prettify<Partial<PromptGroupAwaitedReturn<T>>> }) => void;
+	onCancel?: (opts: { results: Prettify<Partial<T>> }) => void;
 }
 
 type Prettify<T> = {
@@ -862,7 +858,7 @@ type Prettify<T> = {
 
 export type PromptGroup<T> = {
 	[P in keyof T]: (opts: {
-		results: Prettify<Partial<PromptGroupAwaitedReturn<Omit<T, P>>>>;
+		results: Record<PropertyKey, unknown>;
 	}) => undefined | Promise<T[P] | undefined>;
 };
 
@@ -873,7 +869,7 @@ export type PromptGroup<T> = {
 export const group = async <T>(
 	prompts: PromptGroup<T>,
 	opts?: PromptGroupOptions<T>
-): Promise<Prettify<PromptGroupAwaitedReturn<T>>> => {
+): Promise<Prettify<T>> => {
 	const results = {} as any;
 	const promptNames = Object.keys(prompts);
 


43081j:
some examples:

```ts
// Correct return type, correct `results` type in 2nd function
group({
  foo: () => Promise.resolve(303),
  bar: ({results}) => Promise.resolve(808)
});

// Incorrect return type of {}, incorrect `results` type in both functions of {}
group({
  foo: ({results}) => Promise.resolve(303),
  bar: ({results}) => Promise.resolve(808)
});
```

MacFJA:
Another approach is using class: [Typescript Playground](https://www.typescriptlang.org/play/?#code/MYewdgzgLgBAwgQQHJwKIBkD6BlAmgWQCEB5dGAXhmwE8BbAIxABsAKASgCgOpqAHAUxgBpMCADuYAOIAnEAFde2KP14AeACoAaGADUAfBRgsQvKBABcMAN7T+EOUzOX1AXzYUDABVm0AlhH5VfW4+QRl5RWU1dQNKEXEpWQUlFQ1tOTAAa1EJPRCBGG9+KChfADNqDVjrDgBIAG1PGF8wGEz+ahAymHUAXWdG3oBuDhcYADJrFxGOYCYAQwgIGHCFDRh+AA9lMAATZYAlflBpXdVoaRaAc3SsnLBqqxcDKw4Yd5heS4A3eeUYaAqCwwI4nM4Xa7aVaRVIxR4uN4fL6+X7-Wz2RzAzzzaSleZMKqGJ6I97zXZnIQbbb8PbLCFgG66PQsMDzWj8SxCbTzYClcCc+7QlLRbT6NiWaHrSZWGD1TLNVpCfq6GDPGofDUwWxQOTSVrQgB0ZN26kWmRYUAAFv5tKz2dzeb5wJwNQiNYtqGBgFTjnJlMYwHB5l7+ExLMZTMCZeiHE5Cji8QSYqr3OQDN8QL5doZ2B4YBldvwyi1+LtxYUfP5AkUSuVKnD1Zr3mUQNIWKBILA7YIWjArf4DYDeBB3K8m02O9AtXZY4Z5mJ5r5YP2IIOohB6t3eixozPMZYVwaY5iYIsem4RuPNeUjMfYOQH-BkGgsHgiKRRySr5rwEGQ6wbD3ONDzvCA3C-b8Pm1XVWieU9lnUCDxzdSC+2tVdQMMYh6AAK2OKAjSWXwrjAC10KPICIG0KxNzZfhlTvcCrxQptoL1NCB0ws9ENdL9kVRQRoD+XxvWNU0IEyVRUE2fxSgZKkdn2EFjlbcEoEuBlbmyBI9G0SktkUul1MhJkWC2WTrglJI1GkiyGV0mBu05B0+TAAUEiFKIpJk6ATLFRsJ3AKcDJpQts0oMB+DEFZrO8uyrgmaxZXlXslUsHRVWZF1xxCvZSzXIEiRgA0SvM3yGQK4dtFo9llR5VzVSQ942NaXKwq-BEEVmILYCuayc1TAxIui6F2C4DgAHoJpgABaOb5oWxbFpgABVCB5iuQQtjZXgmEEJaFq4SdYG+Qw+oiMapo1AA9AB+L8jXJFgAHIWxAZ7tFzNMKxAPwAgoiBmG+fgWAARgABjYbL3ke3YXvoHEPqMQafr+-gAaBkHntBgAmZ6oYe40XstUMmHez7AIxMw3DzMdx1Ag0EekJqYCuzU7q4K8Gbelm2Zu+6WYZkmmDJ3npvZ+6ryupAQFgahihgQti0i3ZtEBnpsDgGBQFoXbF1aeZGD9ZooEF4oYNRqsMaYYGLWkOR+GhlMHq2X1-R3acqeWGnvrp1jKMZnExabDmpem1AeUtDY9vZMBYC6T3Y2WS15mBtD+F8aQ+1CGBW3zPLldLFnJ2YdH+GkWQ22e4go9EABCJHdy953XSdg0rRpFhlfxI5m994PQ-HEu9oNMmrh3bumF7pOmPp832Mn6fMQ6zgOCAA)

<details>
<summary>Usage screenshot</summary>

image

</details>


With this: 
- each step only knows previous steps result
- the "on cancel" action has a partial of the steps' type,
- the resulting type is the complete type

---

But the DX changes, it's not just a simple function with 2 parameters, but a class instance with multiple functions calls

43081j:
Yes that pretty much works just because of the method chaining. But I don't think we can change the dx so drastically and inconsistently with the other functions 

So we're still left with "weak types inside, but correct everywhere" or "strong types everywhere, but incorrect in most situations"

natemoo-re:
Hmm I'm still not sure where I fall on this... we are looking ahead to a stable v1, so I think it would be okay to introduce a new approach for grouping prompts and deprecate `group()` until removal in v2. I know we're on a v0, so it's technically okay to rip the band-aid off, but that doesn't sit well with me.

43081j:
that makes sense to me

introduce a new interface we can more easily type too and just drop the old one in 2.x

we can have a think about the shape of that on discord

MacFJA:
Related to #139 (for the chaining implementation)

43081j:
closing since this is old and will probably be solved another way

PR #246, opened 2025-03-09
feat!: move to esm-only

43081j:
Migrates to emitting only ES modules now that `require(esm)` is supported in all LTS versions of node other than the soon-to-die-off 18.x.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -18,4 +18,5 @@ jobs:
           node-version: 20
           cache: "pnpm"
       - run: pnpm install
+      - run: pnpm build
       - run: pnpm type-check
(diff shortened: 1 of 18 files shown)


.github/workflows/lint.yml
```diff
@@ -18,4 +18,5 @@ jobs:
           node-version: 20
           cache: "pnpm"
       - run: pnpm install
+      - run: pnpm build
```
43081j, line 21: I'm guessing this worked before because we used `bundler` resolution which basically tells typescript not to care so much about unresolved imports

natemoo-re (approved):
LGTM! Just going to hold off on merging until https://github.com/bombshell-dev/clack/pull/250 lands.

PR #247, opened 2025-03-10
feat(@clack/prompts): custom spinner indicator support

Cygnusfear:
# Custom Spinner Support
Great work on clack! I was eager to have a custom spinner option so I figured I could contribute back to the project. Happy to adjust the implementation if not conform requirements.

Supports a 'custom' indicator on the spinner for settings custom frames and speed:
```ts
export const customSpinner = spinner({
	indicator: "custom",
	custom: {
		frames: ["🦨", "🐎", "🌵", "🤠", "💘", "🧨", "🌮", "🥃", "🏜️"],
		speed: 500,
	},
});
```

### DIFF
--- a/packages/prompts/src/spinner.ts
+++ b/packages/prompts/src/spinner.ts
@@ -12,7 +12,7 @@ import {
 } from './common.js';
 
 export interface SpinnerOptions extends CommonOptions {
-	indicator?: 'dots' | 'timer' | "custom";
+	indicator?: 'dots' | 'timer';
 	onCancel?: () => void;
 	cancelMessage?: string;
 	errorMessage?: string;
@@ -33,10 +33,10 @@ export const spinner = ({
 	output = process.stdout,
 	cancelMessage,
 	errorMessage,
-	...props
+	frames = unicode ? ['◒', '◐', '◓', '◑'] : ['•', 'o', 'O', '0'],
+	delay = unicode ? 80 : 120,
 }: SpinnerOptions = {}): SpinnerResult => {
-	const frames = props.frames ? props.frames : unicode ?  ['◒', '◐', '◓', '◑'] : ['•', 'o', 'O', '0'];
-	const delay = props.delay ? props.delay : unicode ? 80 : 120;
+
 	const isCI = isCIFn();
 
 	let unblock: () => void;
--- a/packages/prompts/src/spinner.ts
+++ b/packages/prompts/src/spinner.ts
@@ -12,10 +12,12 @@ import {
 } from './common.js';
 
 export interface SpinnerOptions extends CommonOptions {
-	indicator?: 'dots' | 'timer';
+	indicator?: 'dots' | 'timer' | "custom";
 	onCancel?: () => void;
 	cancelMessage?: string;
 	errorMessage?: string;
+	frames?: string[];
+	delay?: number;
 }
 
 export interface SpinnerResult {
@@ -31,9 +33,10 @@ export const spinner = ({
 	output = process.stdout,
 	cancelMessage,
 	errorMessage,
+	...props
 }: SpinnerOptions = {}): SpinnerResult => {
-	const frames = unicode ? ['◒', '◐', '◓', '◑'] : ['•', 'o', 'O', '0'];
-	const delay = unicode ? 80 : 120;
+	const frames = props.frames ? props.frames : unicode ?  ['◒', '◐', '◓', '◑'] : ['•', 'o', 'O', '0'];
+	const delay = props.delay ? props.delay : unicode ? 80 : 120;
 	const isCI = isCIFn();
 
 	let unblock: () => void;
--- a/packages/prompts/src/spinner.ts
+++ b/packages/prompts/src/spinner.ts
@@ -36,7 +36,6 @@ export const spinner = ({
 	frames = unicode ? ['◒', '◐', '◓', '◑'] : ['•', 'o', 'O', '0'],
 	delay = unicode ? 80 : 120,
 }: SpinnerOptions = {}): SpinnerResult => {
-
 	const isCI = isCIFn();
 
 	let unblock: () => void;
--- a/packages/prompts/test/spinner.test.ts
+++ b/packages/prompts/test/spinner.test.ts
@@ -176,6 +176,8 @@ describe.each(['true', 'false'])('spinner (isCI = %s)', (isCI) => {
 			for (let i = 0; i < 4; i++) {
 				vi.advanceTimersByTime(200);
 			}
+			
+			result.stop();
 
 			expect(output.buffer).toMatchSnapshot();
 		});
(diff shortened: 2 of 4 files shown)


43081j:
could you catch up from main? there will be big conflicts since i split the modules into individual files 😬 

i'd also rework this slightly to just have a top level `frames` option and `delay` option, rather than `custom`

Cygnusfear:
@43081j for sure!

Caught up and flattened custom into the regular options!

packages/prompts/src/spinner.ts
```diff
@@ -30,9 +32,10 @@ export const spinner = ({
 	output = process.stdout,
 	cancelMessage,
 	errorMessage,
+	...props
 }: SpinnerOptions = {}): SpinnerResult => {
-	const frames = unicode ? ['◒', '◐', '◓', '◑'] : ['•', 'o', 'O', '0'];
-	const delay = unicode ? 80 : 120;
+	const frames = props.frames ? props.frames : unicode ?  ['◒', '◐', '◓', '◑'] : ['•', 'o', 'O', '0'];
+	const delay = props.delay ? props.delay : unicode ? 80 : 120;
```
43081j, line 38: these could be moved to the parameters on line 32 (after `output`):

```ts
frames = unicode ? /* ... */,
delay = unicode ? 80 : 120
```

43081j:
looks good to me, but can you add some tests?

we need to test the two individually:

- custom `frames`
- custom `delay`

you can assert the delay worked by using vitest's fake timers (we already use them in spinner tests)

packages/prompts/src/spinner.ts
```diff
@@ -11,10 +11,12 @@ import {
 } from './common.js';
 
 export interface SpinnerOptions extends CommonOptions {
-	indicator?: 'dots' | 'timer';
+	indicator?: 'dots' | 'timer' | "custom";
```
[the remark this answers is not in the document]
Cygnusfear: ah got it, removed it

packages/prompts/src/spinner.ts
Cygnusfear: yes that's much nicer, moved it

Cygnusfear:
@43081j implemented comments and added tests for spinner & delay, I think I'm doing something wrong but I can't figure out where the issue is-- vitest is reporting these errors after I implement any arbitrary describe/test block:
```bash
(node:39957) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 end listeners added to [ReadableWorkerStdio]. Use emitter.setMaxListeners() to increase limit
(node:39957) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 unhandledRejection listeners added to [process]. Use emitter.setMaxListeners() to increase limit
(node:39957) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 error listeners added to [ReadableWorkerStdio]. Use emitter.setMaxListeners() to increase limit
(node:39957) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 keypress listeners added to [ReadableWorkerStdio]. Use emitter.setMaxListeners() to increase limit
```

packages/prompts/test/spinner.test.ts
```diff
@@ -149,6 +149,31 @@ describe.each(['true', 'false'])('spinner (isCI = %s)', (isCI) => {
 		});
 	});
 
+	describe('indicator customization', () => {
+		test('custom frames', () => {
+			const result = prompts.spinner({ output, frames: ['🐴', '🦋', '🐙', '🐶'] });
+
+			result.start();
+
+			vi.advanceTimersByTime(80);
```
43081j, line 158: if you do this in a loop, the snapshot will contain all 4 frames:

```ts
for (let i = 0; i < 4; i++) {
  vi.advanceTimersByTime(80);
}
```

or `80 * 4`. i think i prefer a loop to make it clear its for the 4 frames

43081j:
that will be because you need to `stop()` your spinner in the test, before asserting the buffer

43081j:
if you catch up from main, you'll get conflicts in your snapshots i think. just make sure locally you `pnpm i` then re-run your tests and update the snapshots

Cygnusfear:
Ah nice, vitest was updated!

Caught up & added frame for loop and `result.stop()`. No snapshot conflicts afaict.

43081j:
seems odd the snapshots didn't update 🤔 

ANSI codes get serialised now into human-readable strings, so it definitely should've updated them

43081j:
ah it looks like you still need to catch up from `main` (github says 2 commits behind, which makes sense as thats where the snapshot change is)

Cygnusfear:
> ah it looks like you still need to catch up from `main` (github says 2 commits behind, which makes sense as thats where the snapshot change is)

my bad, I think I was behind on main updates, woops- did another catch up and indeed new snapshots with ANSI codes replaced

packages/prompts/test/spinner.test.ts
```diff
@@ -145,6 +145,38 @@ describe.each(['true', 'false'])('spinner (isCI = %s)', (isCI) => {
...
+			const result = prompts.spinner({ output, delay: 200 });
+
+			result.start();
+
+			// there are 4 frames
+			for (let i = 0; i < 4; i++) {
+				vi.advanceTimersByTime(200);
+			}
+
 			expect(output.buffer).toMatchSnapshot();
```
43081j, line 180: this one needs a `result.stop()` too it seems

Cygnusfear: will add, will it make sense to add it to earlier describes as well? 

i.e. line `'renders frames at interval'`, line 47?

43081j: you're right. it should be in all of them. otherwise, the cursor will remain hidden after tests run, most likely

Cygnusfear: ok, tried doing this- was a bit hesitant at first..
 
I've put `result.stop()` in the most appropriate positions afaict. Updated snaps to not include the `◇` success character that would appear when firing `stop` before `expect`.

I've refrained from adding it to where the processEmitter is used to terminate the process.

I am assuming now not stopping the result in earlier tests did result in the EventEmitters staying alive & after 10+ tests (not closed) this would result in a warning.

43081j: The stop call should happen before the assertion (the expect) 

Otherwise, if a test fails, the spinner will continue (terminal cursor will be hidden) 

All those tests are testing things like delay and such so it doesn't matter if adding a stop call changes the final printed line

We can have specific tests for the different exit cases

Cygnusfear:
@43081j 
Caught up and moved the stop calls in front of assertions, updated snapshots

43081j (approved):
looks good!

we just need a changeset now and seems good to merge after that (you can run `npx changeset`)

Cygnusfear:
@43081j 
Pushed changeset! Let me know if that looks ok to you 🙏

43081j:
will release it as part of the next 1.0 pre-release (nate may backport it if needed though)

PR #249, opened 2025-03-18
docs: Initial documentation structure and baseline content

dreyfus92:
this is an attempt to work on the docs for clack 🫡


docs/src/content/docs/guides/examples.mdx
```diff
@@ -0,0 +1,308 @@
...
+
+### Simple Text Input
+
+```ts twoslash
+import { text } from '@clack/prompts';
+
+const name = await text({
+  message: 'What is your name?',
+  placeholder: 'John Doe',
+}) as string;
```
43081j, line 18: is it not a string by default? what does `text` resolve to?

43081j:
looks good to me

there's a few places where you cast in examples and i suspect you don't need to 👀 

but other than that this is great

docs/src/content/docs/guides/examples.mdx
43081j: it is, but if you use `isCancel`, it should narrow to `string` i think

dreyfus92: i had to force the type cast, it wasn't able to infer it for some reason 😭

docs/src/content/docs/guides/examples.mdx
```diff
@@ -0,0 +1,308 @@
...
+
+  if (isCancel(features)) {
+    console.log('Operation cancelled');
+    process.exit(0);
+  }
+
+  return {
+    port: Number(port),
+    host,
+    mode: String(mode) as 'development' | 'production',
```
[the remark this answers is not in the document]
dreyfus92: same issue here 🫠

dreyfus92:
closing in favor of moving the docs to it's own repo 😁

PR #253, opened 2025-03-28
refactor: remove duplicated types

bluwy:
There was `dist/index.d.ts` and `dist/index.d.mts` with the same content. With this PR, only `index.d.mts` is kept. As `"types"` is using `.d.mts`, only [TypeScript 4.7](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#new-file-extensions) and above will work, which I think should be ok considering that it's nearly 3 years ago and we're modernizing the package.

Ideally sticking with `index.d.ts` and removing `index.d.mts` could help with compatibility, but it doesn't seem configurable with unbuild.

### DIFF
--- a/build.preset.ts
+++ b/build.preset.ts
@@ -3,7 +3,7 @@ import { definePreset } from 'unbuild';
 // @see https://github.com/unjs/unbuild
 export default definePreset({
 	clean: true,
-	declaration: true,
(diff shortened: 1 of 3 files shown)


dreyfus92 (approved):
thank you @bluwy 🫡

PR #255, opened 2025-03-29
feat: add `selectableGroups` option to group multi-select

43081j:
Adds a `selectableGroups` boolean flag which can be used to disallow selecting groups.

This basically just skips over them when using up/down, but still renders when they are "selected" (by selecting all the children individually).

### DIFF
--- a/.changeset/free-wasps-decide.md
+++ b/.changeset/free-wasps-decide.md
@@ -1,0 +1,6 @@
+---
+"@clack/prompts": patch
+"@clack/core": patch
+---
+
+Adds a new `selectableGroups` boolean to the group multi-select prompt. Using `selectableGroups: false` will disable the ability to select a top-level group, but still allow every child to be selected individually.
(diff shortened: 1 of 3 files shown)


natemoo-re (approved):
Nice, thank you for the PR!

43081j:
@cokakoala are you able to take a look at this and let me know if it matches what you have in svelte's fork?

the behaviour is basically that we render everything as before, but you can't select groups if `selectableGroups` is `false`.

if you select all the children of a group, it will show as selected. but you can't select it yourself

43081j:
i see a missing part actually

in the svelte fork, they also render a different prefix depending on if the groups are selectable or not. i'll take a look at that 👍

43081j:
ok should be the same now

this is how it looks with `selectableGroups: false` and a full group selected
image

this is how it looks with `selectableGroups: true`
image

dreyfus92 (commented):
james, at first glance it looks good but, i just have a question. would it be necessary to add some tests for this?

43081j:
Definitely

We are missing quite a few tests so I might add them in a standalone PR first. I'll take a look

dreyfus92 (approved):

PR #256, opened 2025-03-29
feat: add `groupSpacing` option to group multi select

43081j:
Adds a `groupSpacing` option.

If set to something more than `0`, it adds spacing between the groups.

This is another svelte port. They have `spacedGroups` but I decided to make it a little more flexible here

### DIFF
--- a/.changeset/happy-parents-explain.md
+++ b/.changeset/happy-parents-explain.md
@@ -2,4 +2,4 @@
 "@clack/prompts": minor
 ---
 
-Added `groupSpacing` to control spacing between multi select groups.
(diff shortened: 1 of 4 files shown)


.changeset/happy-parents-explain.md
```diff
@@ -0,0 +1,5 @@
+---
+"@clack/prompts": minor
+---
+
+Added `groupSpacing` to control spacing between multi select groups.
```
natemoo-re, line 5:
```suggestion
Adds a new `groupSpacing` option to grouped multi-select prompts. If set to an integer greater than 0, it will add that number of new lines between each group.
```

natemoo-re (approved):

PR #263, opened 2025-04-01
fix: use placeholder as value when input is empty

jsparkdev:
Fix a bug where the placeholder value is not used when no value is provided to the `text` function.

before:

image

After:
image


jsparkdev:
The following is an example that uses the code from https://bomb.sh/docs/basics/getting-started/#quick-start directly.

https://github.com/user-attachments/assets/10b92147-012c-4864-94c8-eb53f79a3983

Here is how the Vite CLI works.

https://github.com/user-attachments/assets/bf79be4c-8066-41cd-a648-d50c5883300d

The Astro CLI is:

https://github.com/user-attachments/assets/653c85b7-06ba-429a-960f-4ddd46c03fb3

Actually, it looks like Vite CLI uses the prompts package, but its behavior is different, haha.

43081j:
ok i understand now 👍 

i think we should fix this in the core rather than specifically in the text prompt

basically, if you press `<tab>` instead of enter right away, it should fill the value in from the placeholder

the problem seems to be that if you don't tab it in, and just press enter, it doesn't actually set the value.

so we should probably update the core `prompt` code to set `value` on `submit`. i.e. when submitted, if `value === undefined && placeholder !== undefined`, set the `value = placeholder` or some such thing

jsparkdev:
@jsparkdev Thanks for great review.

I've updated the core to behave as follows when there is no input value:

First and foremost, set the default value as the value.
If the default value is not set, set the placeholder as the value.
If both the default value and the placeholder are absent, set an empty string as the value.

dreyfus92 (commented):
first of all thank you @jsparkdev, for taking the time and taking a look at this. this solution might work in some cases, but i think it might have some limitations that could lead into more issues, like ambiguity with falsy values e.g. if `opts.defaultValue` is an empty string (`''`), it skips to `opts.placeholder`, which might not be the desired behavior if an empty string is a valid default value. also we might face type safety issues, this assumes that `opts.defaulValue` and `opts.placeholder` are either strings or falsy, if the could be other types, the result might be unexpected or inconsistent. what about if we try something else? i haven't tested this myself. cc @43081j

packages/core/src/prompts/text.ts
```diff
@@ -26,7 +26,7 @@ export default class TextPrompt extends Prompt {
 
 		this.on('finalize', () => {
 			if (!this.value) {
-				this.value = opts.defaultValue;
+				this.value = opts.defaultValue || opts.placeholder || '';
```
dreyfus92, line 29:
```suggestion
				this.value = (opts.defaultValue !== undefined && opts.defaultValue !== null) ? opts.defaultValue : (opts.placeholder || '');
```

orochaa: That is a nice catch! By the way, you could achieve the same behavior using the Nullish coalescing operator `??`, it works like the OR operator, but only treats `null` and `undefined` as false values.

`opts.defaultValue ?? opts.placeholder ?? ''`

 So if `opts.defaultValue` is nullish, the verifies if `opts.placeholder` is nullish, if so return `''`, otherwise return the first not nullish value.

43081j:
i don't think it belongs in the text prompt

in `Prompt`, we do this:

https://github.com/bombshell-dev/clack/blob/041e13c8eb2927efe5f2edf9ad370a7f2c0466a3/packages/core/src/prompts/prompt.ts#L197-L202

i feel like we should just check here for `'return'` too. i.e. `if ((char === '\t' || key?.name === 'return') && this.opts.placeholder)`

then later, we handle submission etc because return was pressed

packages/core/src/prompts/text.ts
dreyfus92: i forgot about `nullish coalescing` it should do the work as well 🫣. but james' plan sounds better 😁

jsparkdev: thanks!

jsparkdev:
Hmm... it should be modified as follows:

If there is no input value, write and use the default value.
If there are no input and default value, write and use the placeholder.
If there are no input, default, and placeholder values, use an empty string.

packages/core/src/prompts/prompt.ts
```diff
@@ -194,10 +195,15 @@ export default class Prompt {
 		if (char && (char.toLowerCase() === 'y' || char.toLowerCase() === 'n')) {
 			this.emit('confirm', char.toLowerCase() === 'y');
 		}
-		if (char === '\t' && this.opts.placeholder) {
+		if ((char === '\t' || key?.name === 'return') && this.opts.placeholder) {
 			if (!this.value) {
-				this.rl?.write(this.opts.placeholder);
-				this.emit('value', this.opts.placeholder);
+				if (this.opts.defaultValue) {
```
43081j, line 200: we don't need to add `defaultValue` as the individual prompts already deal with that

at some point it makes sense to refactor so we set it in one place i think ,but we don't need to do it yet

if you can add a test for this too, that'd be super helpful (prompt already has some tests)

jsparkdev:
Would it be a bad idea to update the text prompt? like [this commit](https://github.com/bombshell-dev/clack/pull/263/commits/5cd18ea66abddbbf88986c0a62fff063b4579704)

natemoo-re (approved):
Great to see you here, thanks so much for the PR! 🎉

Issue #207, opened 2024-10-20
[Request] Cancelling a spinner


TrifanBogdan24:
How can I know if an user interrupts a spinner with CTRL-C?

PR #264, opened 2025-04-03
feat(prompts): add cancellation support for spinners

dreyfus92:
Closes: #207

### DIFF
--- a/packages/prompts/src/index.ts
+++ b/packages/prompts/src/index.ts
@@ -852,9 +852,6 @@ export const spinner = ({ indicator = 'dots', onCancel }: SpinnerOptions = {}) =
 		start,
 		stop,
 		message,
-		/**
-		 * Returns true if the spinner was cancelled by the user
-		 */
 		get isCancelled() {
 			return isCancelled;
 		}
--- a/packages/prompts/src/index.ts
+++ b/packages/prompts/src/index.ts
@@ -124,9 +124,9 @@ export const text = (opts: TextOptions) => {
 
 			switch (this.state) {
 				case 'error':
-					return `${title.trim()}\n${color.yellow(
-						S_BAR
-					)}  ${value}\n${color.yellow(S_BAR_END)}  ${color.yellow(this.error)}\n`;
+					return `${title.trim()}\n${color.yellow(S_BAR)}  ${value}\n${color.yellow(
+						S_BAR_END
+					)}  ${color.yellow(this.error)}\n`;
 				case 'submit':
 					return `${title}${color.gray(S_BAR)}  ${color.dim(this.value || opts.placeholder)}`;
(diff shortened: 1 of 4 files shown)


packages/prompts/src/index.ts
```diff
@@ -727,22 +727,28 @@ export const stream = {
...
 	let _prevMessage: string | undefined = undefined;
 	let _origin: number = performance.now();
 
 	const handleExit = (code: number) => {
 		const msg = code > 1 ? 'Something went wrong' : 'Canceled';
+		isCancelled = code === 1;
+		if (isCancelled && typeof onCancel === 'function') {
+			onCancel();
+		}
 		if (isSpinnerActive) stop(msg, code);
```
43081j, line 752: since we will stop it here anyway, maybe we should call `onCancel` _after_ this line

that would mean the spinner has been cancelled, therefore stopped, and we've called `onCancel` to let you know

dreyfus92: hmm, you're right. what if we approach this like:

1. detecting cancellation
2. stopping the spinner 
3. call the onCancel callback

```ts
	const handleExit = (code: number) => {
		const msg = code > 1 ? 'Something went wrong' : 'Canceled';
		isCancelled = code === 1;
		if (isSpinnerActive) {
			stop(msg, code);
			if (isCancelled && typeof onCancel === 'function') {
				onCancel();
			}
		}
	};
```

this would ensure the spinner is properly stopped before the callback is invoked.

what do you think?

43081j: that makes sense to me 👍 

you can't really cancel a spinner you haven't started, so i think you're right

packages/prompts/src/index.ts
```diff
@@ -509,9 +510,9 @@ export const groupMultiselect = <Value>(opts: GroupMultiSelectOptions<Value>) =>
 			return `${color.strikethrough(color.dim(label))}`;
 		}
 		if (state === 'active-selected') {
-			return `${color.dim(prefix)}${color.green(S_CHECKBOX_SELECTED)} ${label} ${
-				option.hint ? color.dim(`(${option.hint})`) : ''
-			}`;
+			return `${color.dim(prefix)}${color.green(
```
43081j, line 513: do you know what reformatted this?

if you run `pnpm format`, does it put it back?

dreyfus92: biome did i believe?

43081j: strange that in main it is the same code but formatted it differently 🤔

dreyfus92: let me try to rerun the cmd, i was dealing with errands 🫣

43081j (approved):

PR #265, opened 2025-04-03
fix: export spinner type and add tests

### LINKED ISSUE #248: [Request] Export type for Spinner (ReturnType<typeof spinner>)

I'd like to be able to refer to a created spinner in code.

```ts
let s; // 👈 What type could go here?

// ...

s = spinner();
```

But, `spinner` right now has just an inferred return type:

```ts
declare const spinner: ({ indicator }?: SpinnerOptions) => {
    start: (msg?: string) => void;
    stop: (msg?: string, code?: number) => void;
    message: (msg?: string) => void;
};
```

...so in order to create a type, you'd have to either rewrite that or use some fancy type reference like:

```ts
type Spinner = ReturnType<typeof spinner>;
```

Could we have a `Spinner` type declared in and exported from `@clack/prompts`, please? 🙂 

🤷 I can't think of any.

**Additional context**

43081j:
Exports a `SpinnerResult` which is the API `spinner` produces.

Also adds initial tests for the `prompts` package.

Notable changes:
- `block` consumes a `Readable` and `Writable` now so we can pass our own in when testing (instead of `process.stdout` etc)
- `spinner` accepts an `output` option, maybe we want to move this into a "base options" one day that all prompts accept

Fixes #248

### DIFF
--- a/.changeset/dirty-papayas-happen.md
+++ b/.changeset/dirty-papayas-happen.md
@@ -2,4 +2,4 @@
 "@clack/prompts": patch
 ---
 
-feat: exposes a new `SpinnerResult` type to describe the return type of `spinner`
+Exposes a new `SpinnerResult` type to describe the return type of `spinner`


.changeset/dirty-papayas-happen.md
```diff
@@ -0,0 +1,5 @@
+---
+"@clack/prompts": patch
+---
+
+feat: exposes a new `SpinnerResult` type to describe the return type of `spinner`
```
dreyfus92, line 5:
```suggestion
Exposes a new `SpinnerResult` type to describe the return type of `spinner`
```

dreyfus92 (commented):
just a nit, aside from it LGTM 😁

dreyfus92 (approved):
thank you James 🫡

PR #266, opened 2025-04-04
test: add tests for `text` prompt

43081j:
Adds initial tests for the `text()` prompt.

Depends on #265


43081j:
the diff is a bit crazy because i moved the spinner tests into a `prompts` block, so everything has been indented one

PR #268, opened 2025-04-04
feat: add `CommonOptions` so all prompts accept a custom output/input

43081j:
Allows all prompts to take a custom `output` or `input` which defaults to `process.stdout` and `process.stdin` respectively.


dreyfus92 (commented):
looks good to me, but you might need to resolve issues for approval.

natemoo-re (approved):
Thanks this is a great improvement!

PR #276, opened 2025-04-06
feat: add taskLog prompt

43081j:
Adds a new taskLog prompt which renders messages then ultimately does one of the following:

- If error, retain the output and render the error message
- If success, clear the output and render the success message

For example:

```ts
const log = prompts.taskLog({
  message: 'some message'
});

log.message('a line of text');
log.message('another line of text');
log.message('multiple\nlines\nof\ntext');

// at this point, it will have rendered like so:
// o some message
// | a line of text
// | another line of text
// ...

log.error('some error');

// if the line above runs, it will render like so:
// x some error
// | a line of text
// | another line of text
// ...

log.success('some success');

// if the line above runs, it will clear and render like so:
// o some success
```

part of bringing svelte CLI back to using clack instead of a fork

### DIFF
--- a/packages/prompts/src/index.ts
+++ b/packages/prompts/src/index.ts
@@ -1028,6 +1028,7 @@ export const tasks = async (tasks: Task[], opts?: CommonOptions) => {
 export interface TaskLogOptions extends CommonOptions {
 	message: string;
 	limit?: number;
+	spacing?: number;
 }
 
 export interface TaskLogMessageOptions {
@@ -1041,10 +1042,14 @@ export const taskLog = (opts: TaskLogOptions) => {
 	const output: Writable = opts.output ?? process.stdout;
 	const columns = getColumns(output);
 	const secondarySymbol = color.dim(S_BAR);
+	const spacing = opts.spacing ?? 1;
+	const barSize = 3;
 
 	output.write(`${secondarySymbol}\n`);
 	output.write(`${color.green(S_STEP_SUBMIT)}  ${opts.message}\n`);
-	output.write(`${secondarySymbol}\n`);
+	for (let i = 0; i < spacing; i++) {
+		output.write(`${secondarySymbol}\n`);
+	}
 
 	let buffer = '';
 	let lastMessageWasRaw = false;
@@ -1057,33 +1062,33 @@ export const taskLog = (opts: TaskLogOptions) => {
 			if (line === '') {
 				return count + 1;
 			}
-			return count + Math.ceil(line.length / columns);
+			return count + Math.ceil((line.length + barSize) / columns);
 		}, 0);
-		const lines = bufferHeight + 1 + (lastMessageWasRaw ? 1 : 0) + (clearTitle ? 3 : 0);
+		const lines = bufferHeight + 1 + (clearTitle ? spacing + 2 : 0);
 		output.write(erase.lines(lines));
 	};
 
 	return {
 		message(msg: string, mopts?: TaskLogMessageOptions) {
 			clear(false);
-			if (mopts?.raw) {
+			if (mopts?.raw && lastMessageWasRaw) {
 				buffer += msg;
 			} else {
-				if (lastMessageWasRaw) {
+				if (buffer !== '') {
 					buffer += '\n';
 				}
-				buffer += `${msg}\n`;
+				buffer += msg;
 			}
 			lastMessageWasRaw = mopts?.raw === true;
 			if (opts.limit !== undefined) {
 				const lines = buffer.split('\n');
-				const linesToRemove = lines.length - opts.limit - (lastMessageWasRaw ? 0 : 1);
+				const linesToRemove = lines.length - opts.limit;
 				if (linesToRemove > 0) {
 					lines.splice(0, linesToRemove);
 				}
 				buffer = lines.join('\n');
 			}
-			log.message(lastMessageWasRaw ? `${buffer}\n` : buffer, {
+			log.message(buffer, {
 				output,
 				secondarySymbol,
 				symbol: secondarySymbol,
@@ -1093,7 +1098,7 @@ export const taskLog = (opts: TaskLogOptions) => {
 		error(message: string): void {
 			clear(true);
 			log.error(message, { output, secondarySymbol, spacing: 1 });
-			log.message(buffer, { output, secondarySymbol, symbol: secondarySymbol, spacing: 1 });
+			log.message(buffer, { output, secondarySymbol, symbol: secondarySymbol, spacing });
 		},
 		success(message: string): void {
 			clear(true);
--- a/packages/prompts/src/index.ts
+++ b/packages/prompts/src/index.ts
@@ -1072,12 +1072,12 @@ export const taskLog = (opts: TaskLogOptions) => {
 		message(msg: string, mopts?: TaskLogMessageOptions) {
 			clear(false);
 			if (mopts?.raw && lastMessageWasRaw) {
-				buffer += msg;
+				buffer += color.dim(msg);
 			} else {
 				if (buffer !== '') {
 					buffer += '\n';
 				}
-				buffer += msg;
+				buffer += color.dim(msg);
 			}
 			lastMessageWasRaw = mopts?.raw === true;
 			if (opts.limit !== undefined) {
--- a/packages/prompts/src/index.ts
+++ b/packages/prompts/src/index.ts
@@ -1037,7 +1037,7 @@ export interface TaskLogMessageOptions {
 }
 
 export interface TaskLogCompletionOptions {
-	renderLog?: boolean;
+	showLog?: boolean;
 }
 
 /**
@@ -1117,7 +1117,7 @@ export const taskLog = (opts: TaskLogOptions) => {
 		error(message: string, opts?: TaskLogCompletionOptions): void {
 			clear(true);
 			log.error(message, { output, secondarySymbol, spacing: 1 });
-			if (opts?.renderLog !== false) {
+			if (opts?.showLog !== false) {
 				renderBuffer();
 			}
 			// clear buffer since error is an end state
@@ -1126,7 +1126,7 @@ export const taskLog = (opts: TaskLogOptions) => {
 		success(message: string, opts?: TaskLogCompletionOptions): void {
 			clear(true);
 			log.success(message, { output, secondarySymbol, spacing: 1 });
-			if (opts?.renderLog === true) {
+			if (opts?.showLog === true) {
 				renderBuffer();
 			}
 			// clear buffer since success is an end state
--- a/packages/prompts/src/index.ts
+++ b/packages/prompts/src/index.ts
@@ -1031,11 +1031,11 @@ export const taskLog = (opts: TaskLogOptions) => {
 	output.write(`${color.green(S_STEP_SUBMIT)}  ${opts.message}\n`);
 	output.write(`${color.dim(S_BAR)}\n`);
 
-	const buffer: string[] = [];
+	let buffer = '';
 	let lastMessageWasRaw = false;
 
 	const clear = (clearTitle: boolean): void => {
-		const bufferHeight = buffer.reduce((count, line) => {
+		const bufferHeight = buffer.split('\n').reduce((count, line) => {
 			return count + Math.ceil(line.length / columns);
 		}, 0);
 		const lines = bufferHeight + 1 + (clearTitle ? 2 : 0);
@@ -1045,20 +1045,19 @@ export const taskLog = (opts: TaskLogOptions) => {
 	return {
 		message(msg: string, mopts?: TaskLogMessageOptions) {
 			clear(false);
-			let lines: string[];
-			if (mopts?.raw && buffer.length > 0 && lastMessageWasRaw) {
-				const lastBuffer = buffer.pop();
-				lines = (lastBuffer + msg).split('\n');
+			if (buffer.length === 0 || (mopts?.raw && lastMessageWasRaw)) {
+				buffer += msg;
 			} else {
-				lines = msg.split('\n');
+				buffer += `\n${msg}`;
 			}
 			lastMessageWasRaw = mopts?.raw === true;
-			for (const line of lines) {
-				buffer.push(line);
-			}
-			const linesToRemove = opts.limit !== undefined ? buffer.length - opts.limit : 0;
-			if (linesToRemove > 0) {
-				buffer.splice(0, linesToRemove);
+			if (opts.limit !== undefined) {
+				const lines = buffer.split('\n');
+				const linesToRemove = lines.length - opts.limit;
+				if (linesToRemove > 0) {
+					lines.splice(0, linesToRemove);
+				}
+				buffer = lines.join('\n');
 			}
 			log.message(buffer, { output, spacing: 0 });
 		},
--- a/packages/prompts/src/index.ts
+++ b/packages/prompts/src/index.ts
@@ -1029,10 +1029,15 @@ export interface TaskLogOptions extends CommonOptions {
 	title: string;
 	limit?: number;
 	spacing?: number;
+	retainLog?: boolean;
 }
 
 export interface TaskLogMessageOptions {
 	raw?: boolean;
+}
+
+export interface TaskLogCompletionOptions {
+	renderLog?: boolean;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


manuel3108 (commented):
Overall this looks perfect, although I'm having trouble integrating this into the svelte cli `sv`. But maybe it's just too late for me.

This is how it's currently implemented:
https://github.com/sveltejs/cli/blob/ccb512d1bb48992dcff9cacadd063085789786fd/packages/cli/utils/package-manager.ts#L43-L69

Currently the following exception is printed:
```
TypeError: a.split is not a function
```
I'm assuming this is related to this pr, but from the code i don't understand exactly where that is happening.

43081j:
Ah I may have to tweak it a little

I feel like I remember chunks from a stream not always being line based for stdout. So we probably need to join them up as a string and split again, rather than assuming each one is a line 

I'll take a look. That doesn't fix your problem but I noticed it when reading the usage you linked

manuel3108:
Actually, I think that could exactly be the problem I'm experiencing.

43081j:
We already export `cancel` actually too. I forgot we had that, so we don't need to change it 👍

43081j:
@manuel3108 (and maybe @natemoo-re review):

I've added a `raw` option:

```ts
log.message('line 0', {raw: true});
log.message('still line 0', {raw: true});
log.message('\nline 1', {raw: true}); // now we render another line because we have an explicit `\n`
```

manuel3108:
Sound's logical, but sadly I'm still getting the same error. Here is a repro: https://stackblitz.com/edit/node-edenh16x?file=index.js

43081j:
that is most likely because the event handlers of `stdout` can receive a `Buffer`

if you call `toString()` on the data first, it should work

though the clearing seems a little buggy since some lines don't get cleared which i'd expect to have been

43081j:
@manuel3108 i fixed the bug which was causing leftover buffer lines

your stackblitz works now if you `toString` the chunks

packages/prompts/src/index.ts
```diff
@@ -995,3 +1024,80 @@ export const tasks = async (tasks: Task[], opts?: CommonOptions) => {
...
+			clear(true);
+			log.error(message, { output, secondarySymbol, spacing: 1 });
+			log.message(buffer, { output, secondarySymbol, symbol: secondarySymbol, spacing: 1 });
+		},
+		success(message: string): void {
+			clear(true);
+			log.success(message, { output, secondarySymbol, spacing: 1 });
+		},
+	};
+};
```
AdrianGonz97, line 1103: the below suggestions tweaks the following:
- removed the spacing between the title and the logs
- errors now renders out the full log

```suggestion
export const taskLog = (opts: TaskLogOptions) => {
	const output: Writable = opts.output ?? process.stdout;
	const columns = getColumns(output);
	const secondarySymbol = color.dim(S_BAR);

	output.write(`${secondarySymbol}\n`);
	output.write(`${color.green(S_STEP_SUBMIT)}  ${opts.message}\n`);

	let buffer = '';
	let lastMessageWasRaw = false;
	const frame = () => {
		if (opts.limit !== undefined) {
			const lines = buffer.split('\n');
			const linesToRemove = lines.length - opts.limit - (lastMessageWasRaw ? 0 : 1);
			if (linesToRemove > 0) {
				lines.splice(0, linesToRemove);
			}
			return lines.join('\n');
		}
		return buffer;
	};

	const clear = (clearTitle: boolean): void => {
		if (buffer.length === 0) {
			return;
		}
		const frameHeight = frame()
			.split('\n')
			.reduce((count, line) => {
				if (line === '') {
					return count + 1;
				}
				return count + Math.ceil(line.length / columns);
			}, 0);
		const lines = frameHeight + (lastMessageWasRaw ? 1 : 0) + (clearTitle ? 1 : 0);
		output.write(erase.lines(lines));
	};

	return {
		message(msg: string, mopts?: TaskLogMessageOptions) {
			clear(false);
			if (mopts?.raw) {
				buffer += msg;
			} else {
				if (lastMessageWasRaw) {
					buffer += '\n';
				}
				buffer += `${msg}\n`;
			}
			lastMessageWasRaw = mopts?.raw === true;
			log.message(frame(), {
				output,
				secondarySymbol,
				symbol: secondarySymbol,
				spacing: 0,
			});
		},
		error(message: string): void {
			clear(true);
			log.error(message, { output, secondarySymbol, spacing: 0 });
			log.message(buffer, {
				output,
				secondarySymbol,
				symbol: secondarySymbol,
				spacing: 0,
			});
		},
		success(message: string): void {
			clear(true);
			log.success(message, { output, secondarySymbol, spacing: 1 });
		},
	};
};
```

packages/prompts/src/index.ts
```diff
@@ -995,3 +1024,80 @@ export const tasks = async (tasks: Task[], opts?: CommonOptions) => {
 		s.stop(result || task.title);
 	}
 };
+
+export interface TaskLogOptions extends CommonOptions {
+	message: string;
```
AdrianGonz97, line 1029: `message` feels a bit too close to the `taskLog.message(...)` method and makes it seem like they are connected. i think it might be clearer if it was just called `title`.
```suggestion
	title: string;
```

AdrianGonz97 (commented):
thanks for your work on this!! 

aside from the suggested changes below, it would also be nice if the rendered logs were dimmed with `colors.dim(..)` so that they'd have some nice contrast from the title

packages/prompts/src/index.ts
43081j: can you show me what exactly you changed as this diff isn't very helpful 😬 

also because some of what you're suggesting is already being worked on, so explaining the problem is better than suggesting specific code

packages/prompts/src/index.ts
43081j: this is on purpose because pretty much every other prompt uses `message` for the primary message

i understand what you're saying, but is that worth breaking consistency?

packages/prompts/src/index.ts
43081j: i will implement these myself:

> * removed the spacing between the title and the logs
> * errors now renders out the full log

the first one is as simple as removing the `write` of the separator after the title

the second isn't as simple as what you've suggested here.

the default should be that we _do not_ retain the entire buffer as otherwise it means higher memory usage for everyone when not everyone needs this behaviour. my plan was already to include a new option which lets you toggle this behaviour, so anyone who does want to retain the entire buffer can do so via an option, but the default is to discard it.

43081j:
> aside from the suggested changes below, it would also be nice if the rendered logs were dimmed with `colors.dim(..)` so that they'd have some nice contrast from the title

this is a bug, it was meant to be `dim` already. so i'll fix that when i get chance 👍

43081j:
two updates:

- text is now dim
- there's a `spacing` option to control the spacing between the title and the output

packages/prompts/src/index.ts
AdrianGonz97: it shouldn't be that inconsistent given that `note`'s and `task`'s also provide `title` props: https://github.com/bombshell-dev/clack/blob/3552578ad867f144e10cf33dbd4d46960cace0c9/packages/prompts/src/index.ts#L642

https://github.com/bombshell-dev/clack/blob/3552578ad867f144e10cf33dbd4d46960cace0c9/packages/prompts/src/index.ts#L973-L987

packages/prompts/src/index.ts
AdrianGonz97: > does that make sense? something like `retainLogOnError: true` but a better name

makes perfect sense!

43081j:
ok @AdrianGonz97 @manuel3108 here's where i ended up:

* a new `retainLog` option to the taskLog prompt
  * if this is `true`, the entire log is retained in case you render the log on completion
* a new `renderLog` option to the `success`/`error` methods
  * if this is `true`, the log is rendered after the success/error message
  * by default, `success` will set this to `false`
  * by default, `error` will set this to `true`

thoughts?

AdrianGonz97:
@43081j 
i love this!!

> i am still wondering if `retainLog` is actually something like `ignoreLimitOnComplete` 😅 names are tough

i think `retainLog` is perfect

---

the only issue i'm seeing now is that the text dimming is only _partially_ working. 

here it is on failures: img

similarly fails during the normal process as well: img

this seemingly has to do with how the `buffer` is derived with the `limit`. im guessing the starting part of the `dim`'s ansi escape code is being chopped off

here's a small repro:
```ts
import { setTimeout } from 'node:timers/promises';
import * as p from '@clack/prompts';

p.intro('some-app');

const t = p.taskLog({ title: 'some title', retainLog: true, limit: 3 });

t.message('foo\nbar\nbaz\nfoobar');
await setTimeout(2000); // small timeout to view it during the regular process
t.error('error!');

p.cancel('ahhh');
```

packages/prompts/src/index.ts
```diff
@@ -679,16 +686,38 @@ export const outro = (message = '', opts?: CommonOptions) => {
...
-		const parts = [`${color.gray(S_BAR)}`];
-		if (message) {
-			const [firstLine, ...lines] = message.split('\n');
-			parts.push(`${symbol}  ${firstLine}`, ...lines.map((ln) => `${color.gray(S_BAR)}  ${ln}`));
+		const parts: string[] = [];
+		for (let i = 0; i < spacing; i++) {
+			parts.push(`${secondarySymbol}`);
+		}
+		const messageParts = Array.isArray(message) ? message : message.split('\n');
+		if (message && messageParts.length > 0) {
```
AdrianGonz97, line 707: not needed
```suggestion
		if (messageParts.length > 0) {
```

packages/prompts/src/index.ts
```diff
@@ -995,3 +1024,113 @@ export const tasks = async (tasks: Task[], opts?: CommonOptions) => {
...
+			log.message(`${fullBuffer}\n${buffer}`, {
+				output,
+				secondarySymbol,
+				symbol: secondarySymbol,
+				spacing,
+			});
+		} else {
+			log.message(buffer, { output, secondarySymbol, symbol: secondarySymbol, spacing });
+		}
+	};
```
AdrianGonz97, line 1088: adds `print` so that each line is properly colored
```suggestion
	const renderBuffer = (): void => {
		if (retainLog === true && fullBuffer.length > 0) {
			print(`${fullBuffer}\n${buffer}`);
		} else {
			print(buffer);
		}
	};
	const print = (msg: string): void => {
		const colored = msg.split('\n').map(color.dim).join('\n');
		log.message(colored, { output, secondarySymbol, symbol: secondarySymbol, spacing });
	};
```

packages/prompts/src/index.ts
```diff
@@ -995,3 +1024,113 @@ export const tasks = async (tasks: Task[], opts?: CommonOptions) => {
...
+
+	return {
+		message(msg: string, mopts?: TaskLogMessageOptions) {
+			clear(false);
+			const dimMessage = color.dim(msg);
+			if ((mopts?.raw !== true || !lastMessageWasRaw) && buffer !== '') {
+				buffer += '\n';
+			}
+			buffer += dimMessage;
+			lastMessageWasRaw = mopts?.raw === true;
```
AdrianGonz97, line 1098: coloring is covered by `print`
```suggestion
			clear(false);
			if ((mopts?.raw !== true || !lastMessageWasRaw) && buffer !== '') {
				buffer += '\n';
			}
			buffer += msg;
			lastMessageWasRaw = mopts?.raw === true;
```

packages/prompts/src/index.ts
```diff
@@ -995,3 +1024,113 @@ export const tasks = async (tasks: Task[], opts?: CommonOptions) => {
...
+					}
+				}
+				buffer = lines.join('\n');
+			}
+			log.message(buffer, {
+				output,
+				secondarySymbol,
+				symbol: secondarySymbol,
+				spacing: 0,
+			});
```
AdrianGonz97, line 1115:
```suggestion
			print(buffer);
```

AdrianGonz97 (commented):
`log.message` ended up being the culprit of the partial coloring. more specifically, the coloring of the symbols of the gutter (`│`). 

as the symbols contained the same ansi escape codes as the message's content (e.g. `\x1B[2m│\x1B[22m  \x1B[2msome message goes here\x1B[22m`), the terminating end of the ansi code was not registering properly in multi-line messages.

for example:
```
\x1B[2m│\x1B[22m  \x1B[2mfoo
\x1B[2m│\x1B[22m  bar
\x1B[2m│\x1B[22m  baz
\x1B[2m│\x1B[22m  \x1B[22m
```

it thinks the sequence terminates on the `\x1B[22m` in the second line. as a result, only the first line of the message is properly colored.

the solution was to have each line wrapped with the proper code:
```
\x1B[2m│\x1B[22m  \x1B[2mfoo\x1B[22m
\x1B[2m│\x1B[22m  \x1B[2mbar\x1B[22m
\x1B[2m│\x1B[22m  \x1B[2mbaz\x1B[22m
```

packages/prompts/src/index.ts
43081j: this one is here so when you `message('')`, it won't render blank lines

43081j:
i had to modify what you had a little since we need the `spacing` to be different

but this should all be good now.

i've also renamed the `renderLog` option to `showLog`, which feels a bit nicer

43081j:
@AdrianGonz97 @manuel3108 hows things looking on your end?

if all is good, im happy to merge this once i get an approval

AdrianGonz97:
For the `taskLog`, this is now working perfectly, thank you!! 

The only thing left that we're missing is a replacement for our [`box`](https://github.com/sveltejs/cli/blob/c94db33f88b16cb32602638d7466c2616e89429b/packages/clack-prompts/index.ts#L605), which is effectively a `note` but without the dimming. Not sure if it's solvable in userland, but I'll give it a try in a few.

43081j:
you could probably just `msg.split('\n').map(color.dim).join('\n')` essentially, and pass that into `note`

but maybe it is worth us having some kind of option for this 🤔  though im not sure `dimmed` makes sense as maybe you want it any colour, or any format. not sure where we'd draw the line there, e.g. is it just a `format: (line) => string` option

AdrianGonz97:
> you could probably just `msg.split('\n').map(color.dim).join('\n')` essentially, and pass that into `note`

tried `p.note(msg.split('\n').map(color.reset).join('\n'), "some title");` but ended up with some weirdness where the last line gets partially dimmed 😅
img

> not sure where we'd draw the line there, e.g. is it just a format: (line) => string option

a `format` option would be lovely!

manuel3108:
Just tried this out in the draft PR for `sv`, and this looks perfect. Thank you so much for your help!

I think the new `format` option for `note` should be the final missing piece to get exactly the same functionality as we have right now.

43081j:
I already opened #284 for that FYI

dreyfus92 (approved):
this is pretty solid james, LGTM 🫡

natemoo-re (approved):
Looks awesome! My only comment is on naming—does `taskLog` effectively communicate what this does? Has something like `subtask()`, `subprocess()`, or just `pipe()` been considered?

Not going to block this on an imperfect name, though. Code looks great! Whatever you decide is great 😊

43081j:
its not necessarily from a sub-process, keep in mind

we may use it just for a windowed log from some arbitrary stream

43081j:
i pushed an update FYI

i noticed in CI each `limit` window gets logged since its just a text log and isn't a TTY (so can't erase/move)

so i've updated `message` to render nothing in CI. this way, when you call `error` or `success`, we will render the whole log (assuming you enabled that option).

packages/prompts/src/index.ts
```diff
@@ -1050,6 +1050,7 @@ export const taskLog = (opts: TaskLogOptions) => {
 	const spacing = opts.spacing ?? 1;
 	const barSize = 3;
 	const retainLog = opts.retainLog === true;
+	const isCI = process.env.CI === 'true';
```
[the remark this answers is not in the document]
43081j: there was a reason for not using `isTTY` i think, before i got involved with clack

maybe @natemoo-re can remember 🤔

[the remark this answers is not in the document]
manuel3108: oh, ok. Didn't know that!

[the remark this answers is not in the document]
43081j: opened #286  to tackle this separately

Issue #209, opened 2024-11-20
Improve spinner's cancel messages


luiguild:
https://github.com/bombshell-dev/clack/blob/fc9d52925341a397a6d1a6f9c6d7c6429f441dfa/packages/prompts/src/index.ts#L649

Hello, first of all, thank you for your lib, I've been use it for a while now, and I really want to help you to improve it if it's the case.

I want to find a way to change some behaviors, in this particular case I want to have better cancel messages. The reason for this is that my CLI is multilingual, and these messages end up breaking my flow. All other messages I can customize without issues or, at the very least, intercept the flow to rewrite the output in the terminal.

I believe a good way to work around this would be to capture the spinner’s exit event using the `isCancel` method, just as we can do with prompts. I understand that the method’s purpose is different for prompts, but I believe this approach would maintain the library’s consistency.

Another idea could be to allow passing this message to the constructor when we instantiate the spinner.

I’m open to helping and submitting a PR with these modifications, but I’d like to know your opinion on this. How do you see this situation?

Once again, thank you.

natemoo-re:
Hey, thanks for opening an issue! I think it would be great to expose more control here for multilingual use cases—seems like a relatively small feature!

Maybe this could be handled similarly to the `updateSettings` function exposed from `@clack/core`? Basically just a global object that users can write to before calling the prompts.

https://github.com/bombshell-dev/clack/blob/edd4425f20732d4494fbd9dd71e5c64114991279/packages/core/src/utils/settings.ts#L24-L33

PR #278, opened 2025-04-07
feat: add customizable spinner cancel and error messages

dreyfus92:
closes: #209


43081j:
we should already have some spinner tests in `index.test.ts` FYI

we don't yet test exits, but we do test everything else. so you'd have to copy your process exit stuff across

as for the change itself, I wonder if we should have some i18n-like concept in here. basically a map of all constant strings so people can override them in general

I'll have a think and re-review soon

dreyfus92:
> we should already have some spinner tests in `index.test.ts` FYI
> 
> we don't yet test exits, but we do test everything else. so you'd have to copy your process exit stuff across
> 
> as for the change itself, I wonder if we should have some i18n-like concept in here. basically a map of all constant strings so people can override them in general

sounds good, i will tinker with those suggestions. thanks james! 😁

packages/prompts/src/index.ts
```diff
@@ -776,10 +778,41 @@ export interface SpinnerResult {
...
 
+/**
+ * Global settings for default messages 
+ */
+const promptsSettings = {
+	messages: {
+		cancel: 'Canceled',
+		error: 'Something went wrong',
+	}
+};
```
natemoo-re, line 789: Could this use the [existing global `settings` object](https://github.com/bombshell-dev/clack/blob/46dc0a4bc1b4d8b5ad20aca628334be97887c959/packages/core/src/utils/settings.ts#L10-L22)? We should already expose an `updateSettings()` function

packages/prompts/src/index.ts
```diff
@@ -794,7 +827,9 @@ export const spinner = ({
 	let _origin: number = performance.now();
 
 	const handleExit = (code: number) => {
-		const msg = code > 1 ? 'Something went wrong' : 'Canceled';
+		const msg = code > 1 
+			? (errorMessage ?? promptsSettings.messages.error) 
+			: (cancelMessage ?? promptsSettings.messages.cancel);
```
natemoo-re, line 832: Nice, I like how this override works!

natemoo-re (commented):
Looks great! My only request would be using the existing `updateSettings()` function unless there's a good argument for a separate API here?

natemoo-re:
As for @43081j's suggestion of i18n support, I think that's a great idea! Maybe we can track that as a follow-up. I haven't really seen localized CLI tooling before but that seems pretty amazing.

.changeset/five-chairs-poke.md
```diff
@@ -0,0 +1,5 @@
+---
+"@clack/prompts": minor
+---
+
+Add support for customizable spinner cancel and error messages. Users can now customize these messages either per spinner instance or globally via the new `updatePromptsSettings` function to support multilingual CLIs.
```
natemoo-re, line 5: Need to update this changeset with the proper example!

packages/prompts/src/index.ts
```diff
@@ -12,13 +12,14 @@ import {
 	TextPrompt,
 	block,
 	isCancel,
+	updateSettings as coreUpdateSettings,
+	type ClackSettings as CoreClackSettings,
```
natemoo-re, line 16: I think it's okay to keep the same export name here?

natemoo-re (approved):
Looks great! Please update the changeset and exports before merging 🙏

.changeset/five-chairs-poke.md
dreyfus92: will do it!

packages/prompts/src/index.ts
dreyfus92: typescript yells

.changeset/five-chairs-poke.md
```diff
@@ -0,0 +1,26 @@
+---
+"@clack/prompts": minor
```
natemoo-re, line 2: needs to include core, too!

packages/prompts/src/index.ts
```diff
@@ -11,14 +12,14 @@ import {
...
 	block,
+	updateSettings as coreUpdateSettings,
 	isCancel,
 } from '@clack/core';
 import isUnicodeSupported from 'is-unicode-supported';
 import color from 'picocolors';
 import { cursor, erase } from 'sisteransi';
 
 export { isCancel } from '@clack/core';
-export { updateSettings, type ClackSettings } from '@clack/core';
```
natemoo-re, line 21: These exports need to come back

natemoo-re (commented):
Last things!

packages/core/src/utils/settings.ts
```diff
@@ -40,14 +64,25 @@ export function updateSettings(updates: ClackSettings) {
 
 		switch (key) {
 			case 'aliases': {
-				for (const alias in value) {
-					if (!Object.hasOwn(value, alias)) continue;
+				const aliases = value as Record<string, Action>;
```
43081j, line 67: you don't need to cast this as it could actually be `undefined`

so you should probably just:

```ts
if (value === undefined) {
  break;
}
```

then typescript will know it is non-nullish later

dreyfus92: if i don't cast it i get red squiggly 
image

dreyfus92: the messages cast does work without the assertion 🫡

dreyfus92: i just read about the typeguard you want me to avoid the typecasting 🫣

43081j: specifically you need to handle when it is `undefined`, because it can be

typescript is presumably telling you that in the error in your screenshot

packages/prompts/src/index.ts
```diff
@@ -12,13 +12,14 @@ import {
...
+	updateSettings,
+	settings
 } from '@clack/core';
 import isUnicodeSupported from 'is-unicode-supported';
 import color from 'picocolors';
 import { cursor, erase } from 'sisteransi';
 
-export { isCancel } from '@clack/core';
-export { updateSettings, type ClackSettings } from '@clack/core';
+export { isCancel, updateSettings, settings } from '@clack/core';
```
natemoo-re, line 22:
```suggestion
export { isCancel, updateSettings, settings, type ClackSettings } from '@clack/core';
```

natemoo-re (approved):

PR #281, opened 2025-04-08
fix: validate peer dependencies in release workflow and update ci:version script

dreyfus92:
tests tests tests

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -44,7 +44,7 @@ jobs:
       - name: Validate dependencies
         if: steps.changesets.outputs.published == 'true'
         run: |
-          pnpm dlx knip --production --no-exit-code --reporter=md
+          pnpm run deps:check
         env:
           NODE_ENV: production
       - name: Publish to jsr.io
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -41,7 +41,12 @@ jobs:
         env:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          
+      - name: Validate peer dependencies
+        if: steps.changesets.outputs.published == 'true'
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -41,7 +41,12 @@ jobs:
         env:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          
+      - name: Validate peer dependencies
+        if: steps.changesets.outputs.published == 'true'
+        run: |
+          pnpm dlx dependency-check --peer-deps-only
```
43081j, line 47: what does `dependency-check` come from? which npm package?

dreyfus92: Here is the [link](https://github.com/dependency-check-team/dependency-check). I just noticed that it was archived yikes 😅. So i guess i will try using `knip`

package.json
```diff
@@ -12,21 +12,23 @@
     "type-check": "biome lint && tsc",
     "test": "pnpm --color -r run test",
     "ci:install": "pnpm install --no-frozen-lockfile",
-    "ci:version": "changeset version && pnpx @qingshaner/jsr-release --allowDirty version --sync",
+    "ci:version": "changeset version && pnpm add -D @pnpm/logger@^5.1.0 @pnpm/worker@^1000.1.3 && pnpx @qingshaner/jsr-release --allowDirty version --sync",
     "ci:publish": "changeset publish",
     "ci:prepublish": "pnpm build",
+    "deps:check": "pnpm dlx knip --production --no-exit-code --reporter=md",
```
43081j, line 18: `--no-exit-code` means it'll always exit with a zero exit code (i.e. it won't fail)

do we actually want it to fail so CI won't publish if knip errors?

dreyfus92: i thought that it was going to fail if removed it 😅

43081j (approved):

PR #284, opened 2025-04-11
feat: add `format` option to `note` prompt

43081j:
This adds a new `format` option for formatting lines in a note.

For example:

```ts
// make all lines cyan
note('message', 'title', {
  format(str) {
    return colors.cyan(str);
  }
});

// don't colour lines at all
note('message', 'title', {
  format(str) {
    return str;
  }
});

// add extra text to each line
note('message', 'title', {
  format(str) {
    return `more text ${str} more text';
  }
});
```

cc @AdrianGonz97

### DIFF
--- a/.changeset/empty-buses-wonder.md
+++ b/.changeset/empty-buses-wonder.md
@@ -1,0 +1,5 @@
+---
+"@clack/prompts": minor
+---
+
+Added `formatter` option to the note prompt to allow formatting of individual lines
--- a/.changeset/empty-buses-wonder.md
+++ b/.changeset/empty-buses-wonder.md
@@ -2,4 +2,4 @@
 "@clack/prompts": minor
 ---
 
-Added `formatter` option to the note prompt to allow formatting of individual lines
+Adds `format` option to the note prompt to allow formatting of individual lines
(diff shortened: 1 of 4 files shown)


manuel3108:
This works perfectly!

AdrianGonz97:
This is really excellent @43081j, thank you!! Being able to further enhance the text by line is also very a nice touch 😄 

One small nit: it might be better to just call it `format` since it's performing an action and would follow the existing pattern of option names. the equivalent being the [`validate`](https://github.com/bombshell-dev/clack/blob/44df9af1cb3081aece1bb43de05f05d1f821d546/packages/core/src/prompts/prompt.ts#L17) option for prompts (i.e. it's `validate` instead of `validator`)

.changeset/empty-buses-wonder.md
```diff
@@ -0,0 +1,5 @@
+---
+"@clack/prompts": minor
+---
+
+Added `formatter` option to the note prompt to allow formatting of individual lines
```
dreyfus92, line 5:
```suggestion
Adds `formatter` option to the note prompt to allow formatting of individual lines
```

dreyfus92 (commented):
just a nit 🫡

43081j:
makes sense

have done both changes

dreyfus92 (approved):
thank you @43081j 🫡

PR #287, opened 2025-04-14
fix(prompts): don't strip dots in spinner stop

### LINKED ISSUE #226: [Bug] Unintended Stripping of Periods from Messages
**Environment**

- OS: macOS
- Node Version: v22.12.0
- Package: @clack/prompts
- Package Version: v0.9.0

**Describe the bug**

The [parseMessage](https://github.com/bombshell-dev/clack/blob/1904e57d5a98a1d82b1cd57b42533677d4f9fd8c/packages/prompts/src/index.ts#L690-L692) function is currently removing all trailing period characters (`.`) from the input message. This behavior is problematic because if a message is meant to end with a period, it's being incorrectly stripped, leading to an unintended alteration of the message.

**To Reproduce**

https://stackblitz.com/edit/node-v6ivn9er?file=index.js

- Run `npm run start`.

**Expected behavior**

Messages that end with a single period should retain this period upon completion of the asynchronous process. Only the periods added for loading indicators should be removed without affecting the original message content.

MacFJA:
Fix #226 

---

~~Update `.editorconfig` to avoid trimming trailing space:~~
- ~~in `.snap` files as they contain "raw" CLI output~~
- ~~in `.md` files as `  `+<kbd>enter</kbd> is translated in `<br>`~~

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,9 +10,3 @@ insert_final_newline = true
 [*.yml]
 indent_style = space
 indent_size = 2
-
-[*.snap]
-trim_trailing_whitespace = false
-
-[*.md]
-trim_trailing_whitespace = false
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,3 +10,9 @@ insert_final_newline = true
 [*.yml]
 indent_style = space
 indent_size = 2
+
+[*.snap]
+trim_trailing_whitespace = false
+
+[*.md]
+trim_trailing_whitespace = false
--- a/packages/prompts/src/__snapshots__/index.test.ts.snap
+++ b/packages/prompts/src/__snapshots__/index.test.ts.snap
@@ -1767,6 +1767,20 @@ exports[`prompts (isCI = false) > spinner > stop > renders message 1`] = `
 ]
 `;
 
+exports[`prompts (isCI = false) > spinner > stop > renders message without removing dots 1`] = `
+[
+  "[?25l",
+  "[90m│[39m
+",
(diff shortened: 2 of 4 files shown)


43081j:
while the snapshot change to editor config is true, the editor config would only apply if someone was hand-editing a snapshot. that isn't something anyone should be doing, only vitest should touch them.

similarly, for markdown, i feel like we're more likely to accidentally have trailing whitespace than want a `<br>`.

can you explain the markdown one a little more? to add a new line, you just write `\n\n`, no?

like this message im writing right now, has an empty line above this line, in order to put it on a new line.

MacFJA:
> can you explain the markdown one a little more? to add a new line, you just write `\n\n`, no?

No, this will add a new paragraph.  (https://www.markdownguide.org/basic-syntax/#line-breaks | https://daringfireball.net/projects/markdown/syntax#p)

MacFJA:
@43081j I revert the `.editorconfig` as the use case for Markdown have yet to occurs (and we can still use `<br>` directly, if needed) ; and for the `.snap`, as you have supposed, I edit the snapshot file by hand (I first wrote a test that don't pass, and then make the correction to make it pass)

I also apply your suggestions (renaming the function to something that reflect the goal of the function, and simplify the call in the `stop` method)

43081j:
i understand now 👍  thanks for explaining that. i wasn't aware there was a way to force a line break rather than a paragraph

43081j (approved):

Issue #203, opened 2024-08-27
[Request] can we support autocomplete to filter list/


BennieMeng:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

natemoo-re:
Yes, I'm working on an autocomplete/filter list component currently!

xiapi233:
Looking forward to this feature.

avil13:
Thank you for your work.
Very much missing the filtering of the list to select the desired item.

natemoo-re:
Closing — autocomplete filtering was added in [@clack/core@1.0.0-alpha.0](https://github.com/bombshell-dev/clack/releases/tag/%40clack%2Fcore%401.0.0-alpha.0) with the new `AutocompletePrompt`, and custom filter support was added in [@clack/prompts@1.0.0-alpha.10](https://github.com/bombshell-dev/clack/releases/tag/%40clack%2Fprompts%401.0.0-alpha.10).

PR #288, opened 2025-04-14
feat(core + prompts): adds autocomplete

dreyfus92:
Related: #203  💣


packages/prompts/src/index.ts
```diff
@@ -1014,3 +1014,482 @@ export const tasks = async (tasks: Task[], opts?: CommonOptions) => {
...
+					
+					// Return the formatted prompt
+					return [
+						title,
+						`${color.cyan(S_BAR)}  ${color.dim('Search:')} ${searchText}${matches}${modeIndicator}`,
+						...noResults,
+						...displayOptions.map(option => `${color.cyan(S_BAR)}  ${option}`),
+						`${color.cyan(S_BAR)}  ${color.dim(instructions.join(' • '))}`,
+						`${color.cyan(S_BAR_END)}`
+					].join('\n');
```
MacFJA, line 1239: The variable `title` already end with a `\n`, so by joining all lines with another `\n` we have an empty line between the title and the rest of the prompt

Either the `\n` need to be removed in the title variable (and then re-added everywhere) or it needs to be trim just here.

```suggestion
					return [
						title.replace(/\n$/, ''),
						`${color.cyan(S_BAR)}  ${color.dim('Search:')} ${searchText}${matches}${modeIndicator}`,
						...noResults,
						...displayOptions.map(option => `${color.cyan(S_BAR)}  ${option}`),
						`${color.cyan(S_BAR)}  ${color.dim(instructions.join(' • '))}`,
						`${color.cyan(S_BAR_END)}`
					].join('\n');
```

packages/prompts/src/index.ts
```diff
@@ -1014,3 +1014,482 @@ export const tasks = async (tasks: Task[], opts?: CommonOptions) => {
...
+	const formatOption = (option: Option<Value>, active: boolean) => {
+		const isSelected = selectedValues.includes(option.value);
+		const label = option.label ?? String(option.value ?? '');
+		const hint = option.hint ? color.dim(` (${option.hint})`) : '';
+		const checkbox = isSelected 
+			? color.green(S_CHECKBOX_SELECTED) 
+			: color.dim(S_CHECKBOX_INACTIVE);
+			
+		if (active) {
+			return `${color.green('›')} ${checkbox} ${label}${hint}`;
```
MacFJA, line 1303: The `multiselect` prompt don't have a symbol (`›`) to indicate the position of the cursor, it only relies on the highlighting of the line.

MacFJA (commented):
IMO the hint always visible feel a bit overwhelming, that add a lot of gray (and informative) text
There is also some consistency issue

image

But I really like the presentation: 
- informative line above and under the content
- the content being indented to separate it from the informative part

packages/core/src/prompts/autocomplete.ts
```diff
@@ -0,0 +1,96 @@
...
+        case 'down':
+          this.cursor = this.cursor === this.filteredOptions.length - 1 ? 0 : this.cursor + 1;
+          break;
+      }
+      this.changeValue();
+    });
+
+    // Update filtered options when input changes
+    this.on('value', (value) => {
+      this.value = value;
```
natemoo-re, line 86: I think this needs to account for `space` being both the `select` and potentially part of the search. Right now there's a bug if you switch between navigation and filtering.

packages/prompts/src/index.ts
```diff
@@ -1014,3 +1014,482 @@ export const tasks = async (tasks: Task[], opts?: CommonOptions) => {
...
+					
+					// Update selected value
+					if (filtered.length > 0) {
+						selectedValue = filtered[cursor].value;
+					}
+				});
+				
+				// Register input change handler to update filtering
+				this.on('value', () => {
+					// Only update input when not in navigation mode
```
natemoo-re, line 1139: This doesn't seem to be working! Pressing `space` will still add to the filter value.

packages/prompts/src/index.ts
```diff
@@ -1014,3 +1014,482 @@ export const tasks = async (tasks: Task[], opts?: CommonOptions) => {
...
+					
+					// Render options with selection
+					const displayOptions = filtered.length === 0 
+						? [] 
+						: limitOptions({
+							cursor: cursor,
+							options: filtered,
+							style: (option, active) => {
+								const label = getLabel(option);
+								const hint = option.hint ? color.dim(` (${option.hint})`) : '';
```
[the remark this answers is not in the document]
natemoo-re: I like this suggestion! Think it would help with the visual noise

packages/prompts/src/index.ts
```diff
@@ -1014,3 +1014,482 @@ export const tasks = async (tasks: Task[], opts?: CommonOptions) => {
...
+				
+				case 'cancel': {
+					return `${title}${color.gray(S_BAR)}  ${color.strikethrough(color.dim(this.value || ''))}`;
+				}
+				
+				default: {
+					// Mode indicator (for debugging)
+					const modeIndicator = isNavigating 
+						? color.yellow(' [navigation]') 
+						: '';
```
natemoo-re, line 1192: Maybe we could highlight which mode is active in the footer? `up/down to select` could be white when someone is navigating, `type to search` could be white when someone is in filter mode

packages/prompts/src/index.ts
```diff
@@ -1014,3 +1014,482 @@ export const tasks = async (tasks: Task[], opts?: CommonOptions) => {
...
+						? [`${color.dim('Space:')} select, ${color.dim('Enter:')} confirm, ${color.dim('Type:')} exit navigation`]
+						: [`${color.dim('Type')} to filter, ${color.dim('↑/↓')} to navigate`];
+					
+					// No results message
+					const noResults = filtered.length === 0 && input
+						? [`${color.cyan(S_BAR)}  ${color.yellow('No matches found')}`]
+						: [];
+					
+					// Get limited options for display
+					const displayOptions = limitOptions({
```
natemoo-re, line 1471: Could we update `limitOptions` to accept formatting for the overflow? It'd be nice to indent the `...` with two spaces for this.

natemoo-re (commented):
Looks really awesome! A few minor points on the visual feedback, also noticing a bug with an empty line at the top and some issues switching from multiselect to filtering.

Screenshot 2025-04-18 at 8 39 13 AM

packages/prompts/src/autocomplete.ts
```diff
@@ -0,0 +1,488 @@
...
+		const label = option.label ?? String(option.value ?? '');
+		const hint = option.hint ? color.dim(` (${option.hint})`) : '';
+		const checkbox = isSelected 
+			? color.green(S_CHECKBOX_SELECTED) 
+			: color.dim(S_CHECKBOX_INACTIVE);
+
+		if (active) {
+			return `${color.green('›')} ${checkbox} ${label}${hint}`;
+        }
+		return `${color.dim(' ')} ${checkbox} ${color.dim(label)}${hint}`;
```
MacFJA, line 297:
```suggestion
		return `${color.dim(' ')} ${checkbox} ${color.dim(label)}${isSelected ? hint : ''}`;
```

MacFJA (commented):
I think the best option to handle switching between **navigation** and **filtering** mode is to expose the `_track` property of the `Prompt` class.

In **navigation** `_track` is `false`, and `true` when **filtering**.

---

The `value` event is only triggered when `_track` is `true`,
so if `_track` is `false` we don't have to worry about <kbd>space</kbd> being added in the value.
But in exchange, we have to readd the key that change the mode from **navigation** to **filtering**

MacFJA:
This PR will replace #124

packages/core/src/prompts/autocomplete.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  private changeValue() {
+    if (this.filteredOptions.length > 0) {
+      // Set the selected option's value
+      this.selectedValue = this._value.value;
+    }
+  }
+
+  // Store both the search input and the selected value
+  public selectedValue: any;
```
43081j, line 54: this could probably be a getter, and should use `T['value']` i imagine

```suggestion
  public get selectedValue(): T['value'] | undefined {
    return this.filteredOptions[this.cursor]?.value;
  }
```

packages/core/src/prompts/autocomplete.ts
```diff
@@ -0,0 +1,134 @@
...
+      }
+      this.changeValue();
+    });
+
+    // Update filtered options when input changes
+    this.on('value', (value) => {
+      // Check if we need to ignore a space
+      if (this.ignoreNextSpace && value?.endsWith(' ')) {
+        // Remove the space and reset the flag
+        this.value = value.replace(/\s+$/, '');
```
43081j, line 112: you can probably use `value.trimEnd()` instead

packages/prompts/src/autocomplete.ts
```diff
@@ -0,0 +1,287 @@
...
+							: limitOptions({
+									cursor: this.cursor,
+									options: this.filteredOptions,
+									style: (option, active) => {
+										const label = getLabel(option);
+										const hint = option.hint ? color.dim(` (${option.hint})`) : '';
+
+										return active
+											? `${color.green(S_RADIO_ACTIVE)} ${label}${hint}`
+											: `${color.dim(S_RADIO_INACTIVE)} ${color.dim(label)}${hint}`;
```
43081j, line 120: @MacFJA suggested this shouldn't have a `hint` since it isn't the active option

probably is right (the comment was lost in the refactor)

packages/prompts/src/autocomplete.ts
```diff
@@ -0,0 +1,287 @@
...
+	const formatOption = (option: Option<Value>, active: boolean, selectedValues: Value[]) => {
+		const isSelected = selectedValues.includes(option.value);
+		const label = option.label ?? String(option.value ?? '');
+		const hint = option.hint ? color.dim(` (${option.hint})`) : '';
+		const checkbox = isSelected ? color.green(S_CHECKBOX_SELECTED) : color.dim(S_CHECKBOX_INACTIVE);
+
+		if (active) {
+			return `${color.green('›')} ${checkbox} ${label}${hint}`;
+		}
+		return `${color.dim(' ')} ${checkbox} ${color.dim(label)}${hint}`;
```
43081j, line 202: another lost @MacFJA comment. that we should only show the hint here if the option is focused

something like:

```ts
const hint = option.hint && option.value === this.focusedValue ? color.dim(` (${option.hint})`) : '';
```

MacFJA, line 202: The `multiselect` prompt don't rely on `›` (which is a Unicode (1.1) char, so it should be handled by the `unicodeOr` function) to indicate which line is focused.

The `multiselect` use the fact that only the focused line is not `dim`

packages/prompts/src/autocomplete.ts
```diff
@@ -0,0 +1,298 @@
...
+	 * Placeholder text to display when no input is provided.
+	 */
+	placeholder?: string;
+}
+
+export const autocomplete = <Value>(opts: AutocompleteOptions<Value>) => {
+	const prompt = new AutocompletePrompt<Option<Value>>({
+		options: opts.options,
+		placeholder: opts.placeholder,
+		initialValue: opts.initialValue ? [{ value: opts.initialValue }] : undefined,
```
43081j, line 71: this was right before

the `initialValue` is the value, not the option. it was meant to be `[opts.initialValue]`

presumably this is also why you added `Option<Value>`, because it was inferring the right type before but adding this broke it (correctly)

packages/prompts/src/autocomplete.ts
```diff
@@ -0,0 +1,298 @@
...
+		multiple: true,
+		filter: (search, opt) => {
+			return getFilteredOption(search, opt);
+		},
+		placeholder: opts.placeholder,
+		initialValue: opts.initialValues,
+		input: opts.input,
+		output: opts.output,
+		render() {
+			const formatOption = (option: Option<Value>, active: boolean, selectedValues: Value[]) => {
```
43081j, line 211: you moved this into `render` so you can access `this.focusedValue` i imagine

we should just add another parameter, we shouldn't have to create this function every render

```ts
const formatOption = (option: Option<Value>, active: boolean, selectedValues: Value[], focusedValue: Value | undefined) => {
```

packages/prompts/src/autocomplete.ts
```diff
@@ -208,11 +205,16 @@ export const autocompleteMultiselect = <Value>(opts: AutocompleteMultiSelectOpti
 		input: opts.input,
 		output: opts.output,
 		render() {
-			const formatOption = (option: Option<Value>, active: boolean, selectedValues: Value[]) => {
+			const formatOption = (
```
43081j, line 208: this can be moved back outside of the `render` method now

packages/core/src/prompts/prompt.ts
```diff
@@ -223,9 +231,11 @@ export default class Prompt {
 			}
 		}
 
+		// Only process as cancel if the key wasn't already handled
```
43081j, line 234: this comment no longer applies i think

packages/prompts/src/limit-options.ts
```diff
@@ -8,12 +8,18 @@ export interface LimitOptionsParams<TOption> extends CommonOptions {
...
+	 * Defaults to '    ...' if not provided.
+	 */
+	overflowFormat?: string;
 }
 
 export const limitOptions = <TOption>(params: LimitOptionsParams<TOption>): string[] => {
 	const { cursor, options, style } = params;
 	const output: Writable = params.output ?? process.stdout;
 	const rows = output instanceof WriteStream && output.rows !== undefined ? output.rows : 10;
+	const overflowFormat = params.overflowFormat ?? color.dim('    ...');
```
43081j, line 22: did we ever use this? it looks like the only places we set it, we use the default

packages/prompts/test/autocomplete.test.ts
```diff
@@ -0,0 +1,124 @@
...
+		input = new MockReadable();
+		output = new MockWritable();
+	});
+
+	afterEach(() => {
+		vi.restoreAllMocks();
+	});
+
+	test('renders initial UI with message and instructions', async () => {
+		const result = autocomplete({
```
43081j, line 26: we should probably make sure these get disposed properly

e.g.

```ts
const result = autocomplete(...);

input.emit('key', '', {name: 'return'});
await result;
```

just so we don't end up with loads of dangling promises all over

packages/prompts/test/autocomplete.test.ts
```diff
@@ -0,0 +1,124 @@
...
+		});
+
+		expect(output.buffer).toMatchSnapshot();
+	});
+
+	test('limits displayed options when maxItems is set', async () => {
+		const result = autocomplete({
+			message: 'Select a fruit',
+			options: testOptions,
+			maxItems: 2,
```
43081j, line 52: we force `maxItems` to be a minimum of 5 internally

so this test should probably generate some options and set the `maxItems` to 6 or something

```ts
const options = [];
for (let i = 0; i < 10; i++) {
  options.push({ value: `option ${i}` });
}
```

packages/prompts/test/autocomplete.test.ts
```diff
@@ -0,0 +1,124 @@
...
+			message: 'Select a fruit',
+			options: testOptions,
+			input,
+			output,
+		});
+
+		expect(output.buffer).toMatchSnapshot();
+	});
+
+	test('shows placeholder when provided', async () => {
```
43081j, line 36: i don't think we actually render the placeholder

the snapshot shows this

so we should decide either:
- remove this test and autocomplete doesn't support placeholders
- make placeholders work and update this test's snapshot

packages/prompts/src/limit-options.ts
43081j: looking at the snapshots, it seems this changes how some of the existing prompts render
so maybe we should just revert this bit as it seems unnecessary for the autocomplete

packages/prompts/test/autocomplete.test.ts
dreyfus92: i opted for removing the test 🫡

packages/prompts/src/limit-options.ts
43081j: should we maybe revert this @dreyfus92?

unless you can figure out why it exists

43081j: the multi-select prompt had no spaces before, and now has two, because of this

that's what i want to understand. does it look better with the two spaces? why were they added?

43081j (approved):

grabbou:
CleanShot 2025-05-13 at 00 55 06@2x

Feature wise this works awesome! Unfortunately, I am seeing this weird glitch where things appear multiple times. Using both Ghostty and built-in Mac terminal.

dreyfus92:
hey @grabbou can you open an issue please? 😁

PR #290, opened 2025-04-14
feat(prompts): Add progressbar

MacFJA:
Add a new prompt: `progress`

It's a superset of spinner.

The progress bar came in 3 flavor:
- `block` (Unicode: `█`, ASCII: `#`)
- `heavy` (Unicode: `━`, ASCII: `=`) [default value]
- `light` (Unicode: `─`, ASCII: `-`)

The color scheme is the same as the spinner (fill part is magenta, missing part is dim)


packages/prompts/src/index.ts
```diff
@@ -929,6 +929,85 @@ export const spinner = ({
...
+}: ProgressOptions = {}): ProgressResult {
+	const spin = spinner({
+		onCancel,
+		cancelMessage,
+		errorMessage,
+		indicator,
+		output,
+		input,
+	});
+	const progressChar: Record<NonNullable<ProgressOptions['style']>, string> = {
```
43081j, line 964: can we move this outside the function since it is static?

we should probably do the same for spinner frames in a separate PR one day

packages/prompts/src/index.ts
```diff
@@ -929,6 +929,85 @@ export const spinner = ({
...
+			case 'initial':
+			case 'active':
+				return color.magenta;
+			case 'error':
+			case 'cancel':
+				return color.red;
+			case 'submit':
+				return color.green;
+		}
+		return color.magenta;
```
43081j, line 986: typescript will detect this as unreachable since we have checked all possible `State` values

if you use `default` instead, typescript will be happy

```ts
default:
  return color.magenta;
```

packages/prompts/src/index.ts
MacFJA: I also renamed it to match the rest of the static values in the file.

packages/prompts/src/index.ts
```diff
@@ -929,6 +929,85 @@ export const spinner = ({
 	};
 };
 
+export type ProgressOptions = SpinnerOptions & {
```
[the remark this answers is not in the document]
MacFJA: I update the `ProgressResult` to also extends its spinner version

packages/prompts/src/common.ts
```diff
@@ -2,6 +2,7 @@ import type { Readable, Writable } from 'node:stream';
 import type { State } from '@clack/core';
 import isUnicodeSupported from 'is-unicode-supported';
 import color from 'picocolors';
+import type { ProgressOptions } from './progress-bar.js';
```
MacFJA, line 5: I don't really like the fact that `common.ts` is dependent of `progress-bar.(t|j)s`.

But I don't know if best solution exist :thinking: 

----

Maybe the `S_PROGRESS_CHAR` can be left in the `progress-bar` file, but the `s` function need to be exported then.

packages/prompts/src/index.ts
```diff
@@ -4,14 +4,13 @@ export * from './common.js';
 export * from './confirm.js';
 export * from './group-multi-select.js';
 export * from './group.js';
-export * from './index.test.js';
-export * from './index.js';
```
MacFJA, line 8: No need to re-export the current file.

packages/prompts/src/common.ts
43081j: we should just keep them in the progress bar module, since nothing else needs them

if they were shared/"common", they'd be individual constants like the ones above

we can export `s` as long as we give it a real name, like `unicodeFallback` or something

packages/prompts/src/index.test.ts
```diff
@@ -322,6 +322,42 @@ describe.each(['true', 'false'])('prompts (isCI = %s)', (isCI) => {
 				prompts.settings.messages.error = originalErrorMessage;
 			});
 		});
+
+		describe('progress', () => {
```
43081j, line 326: this should be in its own describe block rather than inside the `spinner` one

its ok to repeat the `beforeEach`/`afterEach`

packages/prompts/src/index.test.ts
```diff
@@ -2,9 +2,10 @@ import { EventEmitter, Readable, Writable } from 'node:stream';
 import colors from 'picocolors';
 import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, test, vi } from 'vitest';
 import * as prompts from './index.js';
+import progress from './progress-bar.test.js';
 
 // TODO (43081j): move this into a util?
-class MockWritable extends Writable {
+export class MockWritable extends Writable {
```
43081j, line 8: if you catch up from main, you can discard these changes and import from `test-utils.js`

packages/prompts/src/index.test.ts
```diff
@@ -65,6 +66,8 @@ describe.each(['true', 'false'])('prompts (isCI = %s)', (isCI) => {
 		vi.restoreAllMocks();
 	});
 
+	progress(() => ({ isCI, input, output }));
```
43081j, line 69: what is this line for?

packages/prompts/src/progress-bar.ts
```diff
@@ -0,0 +1,70 @@
...
+	style = 'heavy',
+	max = 100,
+	size = 40,
+	...spinnerOptions
+}: ProgressOptions = {}): ProgressResult {
+	const spin = spinner(spinnerOptions);
+	let value = 0;
+	let previousMessage = '';
+
+	const _max = Math.max(1, max);
```
43081j, line 32: since a name prefixed with `_` usually means "ignore this being unused", maybe we should name these two `max` and `size`. then name the inputs, `userMax` and `userSize` or something.

```ts
export function progress({
  max: userMax = 100,
  size: userSize = 40
```

43081j (approved):

43081j:
looks good to me!

can you add a changeset? i normally just run `npx changeset`

Issue #291, opened 2025-04-15
[Request] Use type stripping for examples


43081j:
We currently use jiti in the examples package in order to run the typescript sources.

In node 23.x, we can run these natively through node (i.e. `node src/spinner.ts` will work).

We could do this now but it means maintainers can only run examples under node 23, which may not be ideal (yet?).

We should at the very least enable `erasableSyntaxOnly` in our repo

PR #292, opened 2025-04-15
chore: enable `erasableSyntaxOnly` in typescript

43081j:
This means we can directly execute our examples one day.

Related to #291


dreyfus92 (approved):
thank you @43081j 🫡

PR #295, opened 2025-04-16
test: split up tests into individual files

43081j:


43081j:
i reviewed this by running this script:

```ts
import {readdir, readFile} from 'node:fs/promises';
import * as path from 'node:path';

const files = await readdir('./packages/prompts/src/__snapshots__');
const oldSnap = await readFile('./foo.snap', 'utf-8');

const oldExports = {};
Function('exports', oldSnap)(oldExports);

for (const file of files) {
	const contents = await readFile(path.join('./packages/prompts/src/__snapshots__', file), 'utf-8');
	const newExports = {};
	Function('exports', contents)(newExports);

	for (const key in newExports) {
		const match = key.match(/^(\w+) \(isCI = (true|false)\)(.+)$/);

		const oldKey = `prompts (isCI = ${match[2]}) > ${match[1]}${match[3]}`;
		if (oldExports[oldKey] !== newExports[key]) {
			console.warn('Whoops!', {oldKey, key});
		}
	}
}
```

all of them were equal, i.e. no `Whoops!` was logged

so snapshots seem good 👍

43081j:
merging to unblock some other PRs 👍

PR #299, opened 2025-04-17
chore: upgrade vitest, tweak knip

43081j:
Upgrades vitest and fixes up some knip stuff.

Summary:

- Examples need their entry/project explicitly set in knip since all files should be considered "exported"
- For `--production` to work, we need to declare non-test files as "production" with a `!` suffix
- vitest 3.x messes with term colours for whatever reason, so we now need FORCE_COLOR=1


packages/core/src/prompts/text.ts
```diff
@@ -1,7 +1,7 @@
 import color from 'picocolors';
 import Prompt, { type PromptOptions } from './prompt.js';
 
-export interface TextOptions extends PromptOptions<TextPrompt> {
+interface TextOptions extends PromptOptions<TextPrompt> {
```
43081j, line 4: these don't get exported from `src/index.ts` currently. we need to export all the `Options` from all the prompts in a separate PR

PR #301, opened 2025-04-17
fix(prompts): Fix styling of taskLog + fix README

MacFJA:


packages/prompts/src/common.ts
```diff
@@ -3,8 +3,15 @@ import type { State } from '@clack/core';
 import isUnicodeSupported from 'is-unicode-supported';
 import color from 'picocolors';
 
-export const unicode = isUnicodeSupported();
-export const unicodeOr = (c: string, fallback: string) => (unicode ? c : fallback);
+export const env = {
```
43081j, line 6: imo these are two unrelated exports

```ts
export const unicode = isUnicodeSupported();
export const isCI = () => process.env.CI === 'true';
```

MacFJA: The advantages of the `env` object is the getter syntax that introduce less changes in the code (instead of `unicode`, it was just `env.unicode`) and it allows to unit test Unicode support (but only for spinner as all other usages are computed once).
And it is the same for `env.isCI`: with the getter syntax it is used as it was a simple property, where in reality it's a function. (Yet again, mainly to help unit testing)

---

But I can move these changes in another PR

packages/prompts/src/task-log.ts
```diff
@@ -26,11 +26,11 @@ export interface TaskLogCompletionOptions {
 export const taskLog = (opts: TaskLogOptions) => {
 	const output: Writable = opts.output ?? process.stdout;
 	const columns = getColumns(output);
-	const secondarySymbol = color.dim(S_BAR);
+	const secondarySymbol = color.gray(S_BAR);
```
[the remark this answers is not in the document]
MacFJA: The `gray` and the `dim` don't render the same.

Everywhere else in clack, it's `gray`, only here it's `dim`.

If you look at the rendering of `npx sv create` you will see that is display is inconsistent.

Before this PR the rendering is this (same content as `examples/basic/task-log.ts`, I just add a log before and after): 
image
image

packages/prompts/src/common.ts
43081j: we don't mock them anywhere though, we mock what they rely on (e.g. `process.env`)

i can't see any functional reason for having them in this `env` object. i'd just export them individually

packages/prompts/src/task-log.ts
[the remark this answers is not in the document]
43081j: i know dim and grey are different. the `taskLog` is mostly ported from svelte, though and they specifically had `dim` in there. presumably so the log would retain colour but be dimmer than the header

are you seeing different somewhere?

packages/prompts/src/common.ts
43081j: in actual fact there'd be fewer changes that way too because all references to `unicode` would stay as they are, no?

43081j: yes, because it is an ES module export like all the others we have

but we don't need to mock it anywhere. were you trying to make it mockable in preparation for new tests?

43081j (approved):

PR #302, opened 2025-04-17
feat(core+prompts): Refactoring of the frame renderer

MacFJA:
Extract the rendering function from `Prompt` class into a reusable function.

The goal is to provide a function to easily renderer a frame (with partial re-rendering).

----

Update `spinner` prompt to use the new function


43081j:
the idea here is good, to clean up some duplicated logic. but i think our prompts are too varied in how they render right now

this "renderer" probably should handle a whole bunch of other behaviours, which may mean it should even be a class, etc

i'm not sure we should land this as it is, and should spend some time reading into how the other prompts render first. otherwise we basically extracted the logic `spinner` needs and not much else.

e.g taskLog has logic for a windowed buffer, which could be an option of a renderer instance.

43081j:
closing this as we're working on a new rendering function so this will conflict/become redundant in the near future

thank you for the time spent though, as this has helped us better understand the way we want to go

PR #305, opened 2025-04-18
feat(prompts): Merge stream and log utilities

### LINKED ISSUE #304: [Request] Merge `log` and `stream`
The `@clack/prompts` package has both `log` and `stream` helpers, but the only difference is that `stream` accepts iterables and `log` accepts strings. Ideally these could be merged into a single API that handles both.

MacFJA:
Closes #304 

---

TODO:
- [ ] Add changeset
- [ ] Update README
- [ ] Update Unit test


43081j:
im going to close this since there's a larger upcoming refactor which might tackle this a different way

PR #308, opened 2025-04-28
core rework/simplification

43081j:
This is a bunch of refactoring on top of the autocomplete prompt which uncovered a whole bunch of possible bugs/awkward code we could fix.


dreyfus92 (approved):
this is so good, thank you @43081j

PR #309, opened 2025-04-28
fix(prompts): Fix dots loader in spinner

MacFJA:
Previously, the dots indicator (the ellipsis) were linked to the symbols used, 4 char (`◒`, `◐`, `◓`, `◑`)

The timing of the dots was one dot every 8 frame (+0.125 every frame, so a integer number every 8 frame) which was OK since we display 3 dots.

But now that the spinner frame char can be changed, this logic don't work anymore.

For example, if you have 10 spinner frames you will have (excluding non-integer number):
```

.
..
...
...
...
...
...
...
...

.
..
...
...
...
```
And if you have only 2 spinner frames :
```

.
..

.
..
```

---

This PR fix this by basing the dots indicator only on itself


43081j (changes requested):
makes sense to me

can you add a changeset?

MacFJA:
> makes sense to me
> 
> can you add a changeset?

Added !
(Sorry, I always forget it :scream: )

43081j (approved):

Issue #315, opened 2025-05-12
[Bug] validation not triggered when using initialValue


mdjastrzebski:
**Environment**

- OS: macOS
- Node Version: v20.12.0
- Package: `@clack/prompts`
- Package Version: 0.10.0

**Describe the bug**
When using `text` and passing `initialValue` and `validate` props the validation is not triggered for incorrect `initialValue`.

**To Reproduce**
https://stackblitz.com/edit/node-hchksvjq?file=index.mjs

- Run the stackblitz

**Expected behavior**
For incorrect initial value the validator should display the validation message. But it does not. User has to press enter first.

Alternatively there should be option to trigger validation.

**Additional Information**

dreyfus92:
Thank you for reporting this @mdjastrzebski 😁

PR #316, opened 2025-05-13
fix: validate initial values

dreyfus92:
When using text prompts with `initialValue` and `validate` props, validation was not being triggered for incorrect initial values. Users had to press enter first to see validation messages.

This fix:
- Validates initial values as soon as they are set
- Shows validation messages immediately for invalid initial values
- Maintains the prompt in error state until a valid value is entered

Added a couple of tests and an example.

Closes: #315

### DIFF
--- a/.changeset/mean-mice-train.md
+++ b/.changeset/mean-mice-train.md
@@ -1,0 +1,5 @@
+---
+"@clack/core": patch
+---
+
+Validates initial values immediately when using text prompts with initialValue and validate props.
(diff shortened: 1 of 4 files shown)


43081j (approved):
this looks good to me

loosely related, i do wonder if we should have some kind of "validate on key press" option one day. currently, we will validate when you press `<return>` and not when you're typing. this PR conflicts with that since the user hasn't pressed return yet, but does make the UX nicer ofc.

MacFJA:
> loosely related, i do wonder if we should have some kind of "validate on key press" option one day. currently, we will validate when you press `<return>` and not when you're typing. this PR conflicts with that since the user hasn't pressed return yet, but does make the UX nicer ofc.

I made a change in this direction in #240

PR #323, opened 2025-05-18
fix: treat placeholders as visual element

43081j:
This changes how we deal with placeholders.

Currently, a placeholder behaves as a visual hint of what we want the user to type. It also behaves as an optional default value, in that you can press `<Enter>` or `<Tab>` to insert it as the value.

This is confusing since it means it is a "partial" default value, and a placeholder at the same time.

This change basically removes the placeholder logic from core, such that the `text` and `autocomplete` prompts use `placeholder` purely as a render/visual hint.

This updated behaviour means we render the placeholder when no value is set, but it can no longer be inserted as a value. To achieve that, you should combine `defaultValue` with `placeholder` (i.e. set them to the same value).

cc @dreyfus92 @natemoo-re

### DIFF
--- a/.changeset/mean-turkeys-help.md
+++ b/.changeset/mean-turkeys-help.md
@@ -1,0 +1,6 @@
+---
+"@clack/prompts": patch
+"@clack/core": patch
+---
+
+Changes `placeholder` to be a visual hint rather than a tabbable value.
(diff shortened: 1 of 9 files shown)


dreyfus92 (approved):
I like this idea, but i don't know if it aligns with nate 😁

43081j:
the original idea seemed to be that `placeholder` was a value you can tab in, like an optional default

but we can achieve this with `defaultValue` a bit more cleanly i think

if you set `defaultValue` in a text prompt, pressing `<enter>` will set the value as that. meanwhile, the `placeholder` is just to show a message until the user types something in

i think this makes more sense than whats in main

dreyfus92 (commented):
I just noticed that the changeset wasn't detected by the bot 😅

manuel3108:
This aligns with what I'm expecting and was used to in 0.x. Additionally, this seem like the cleaner way for me. Strongly approve.

PR #334, opened 2025-05-28
feat: rework values and user input to be separate

### LINKED ISSUE #330: Prompt `value` should only be mutated on finalise
Currently, we have some freaky behaviours in that `value` is repurposed through the lifetime of a prompt.

Let's take autocomplete multi for example. It goes like this:

- `value` is empty
- user types in `foo`
- `value` is now `"foo"`
- user selects the "foo" option with `<TAB>`
- user presses `<RETURN>`
- `value` is now `[ 'foo' ]`

so the type is basically `string | undefined | string[]`

this is pretty confusing and means functions which deal with `value` (like `validate`) have no clue what type the `value` will be

could be a string, an array, an object, whatever. its basically `unknown`

i think we should mutate some internal `value` which represents **user input** (when they type something in), and mutate `value` for the prompt's strongly typed value

43081j:
Fixes #330 

Basically, this introduces a new `userInput` field and event to the core.

The difference is:
- `userInput` tracks what the user has typed in (if `track: true`)
- nothing changes `value` in the core anymore

This means it is the responsibility of each prompt to decide what to do with `value`. They can then call `this._setValue(value)` to set the value and trigger the `value` event.

### An example - text prompt

The text prompt wants to set the value to what the user has typed in. So we can simply listen for `userInput` and set `value` to whatever they have typed. Or we can wait for `finalize` and do it.

### A more complex example - auto complete

The auto complete prompt wants to take user input, but this isn't the value. The value should be the _completed value_.

To achieve this, we can set `value` only once someone selects an option.

This means the `value` is only ever `undefined` or a selected option.

### DIFF
--- a/packages/core/src/prompts/autocomplete.ts
+++ b/packages/core/src/prompts/autocomplete.ts
@@ -44,13 +44,16 @@ function normalisedValue<T>(multiple: boolean, values: T[] | undefined): T | T[]
 	return values[0];
 }
 
-interface AutocompleteOptions<T extends OptionLike> extends PromptOptions<T['value'] | T['value'][], AutocompletePrompt<T>> {
+interface AutocompleteOptions<T extends OptionLike>
+	extends PromptOptions<T['value'] | T['value'][], AutocompletePrompt<T>> {
 	options: T[];
 	filter?: FilterFunction<T>;
 	multiple?: boolean;
 }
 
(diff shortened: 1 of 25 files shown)


dreyfus92 (commented):
i really like this refactor, you're just lacking a changeset and i'm wondering if this changes impact all the examples we've created in the examples dir

PR #335, opened 2025-05-28
feat: rework path, remove suggestion prompt

43081j:
After some thought recently, I came to the conclusion that the `path` and `suggestion` prompts shouldn't exist in their current form.

The `path` prompt is a file selector - but is basically giving a poorer equivalent of the autocompletion UX. Instead of providing two ways to render completions, I think we should settle on the autocompletion prompt and remove the suggestion prompt.

This means the `path` prompt now uses the autocompletion prompt, but with `options` sourced from the file system.

## Notable changes

- Autocomplete prompt can take an options function as well as an options array. This will be called each time we need to query the options set
- Autocomplete prompt now sets its `value` on `<RETURN>` rather than `finalize` event. This allows us to set the `value` before validation occurs, so we can then validate against the completed path
- Core prompt no longer has a `beforePrompt` and instead provides an `initialUserInput` which can be used to set the input without setting the `value`
- Autocomplete prompt can now take a `validate` function and will render validation errors

## Behavioural differences compared to the old `path`

In the old `path`, you could enter a value which doesn't exist as an option and it would be set as the result on submit.

In the new `path`, you **must** select a value. you can no longer submit arbitrary/unknown values

cc @dreyfus92

also @MacFJA you put plenty of work into these two prompts, i know. i think it is sensible to consolidate, though. we can maintain one core prompt instead of two here and reduce the maintainer burden. this will also keep the UX consistent.


dreyfus92 (approved):
LGTM, we might lack to update docs tho for this part, we can add them in the docs repo.

43081j:
@manuel3108 would you be able to try this branch out in svelte? 

It includes the placeholder changes in main that unblock your unforking

43081j:
@dreyfus92 im going to open an issue to track updating the docs, as the original PR didn't add any either

so we are missing a few entries

PR #350, opened 2025-06-22
fix: correctly display deleted message

btea:
When the text method is called, and a message with spaces before and after it is entered in the terminal, ` a `, and then the operation is canceled, you can see that the entered message is deleted, but the space after the message is also removed. The displayed message is inconsistent with the input.

|before|after|
|---|---|
|image|image|

### DIFF
--- a/packages/core/src/prompts/prompt.ts
+++ b/packages/core/src/prompts/prompt.ts
@@ -261,7 +261,7 @@ export default class Prompt<TValue> {
 	}
 
 	private render() {
-		const frame = wrap(this._render(this) ?? '', process.stdout.columns, { hard: true });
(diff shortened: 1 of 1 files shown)


43081j:
you've actually uncovered a bit of a pandora's box here

i think we should be turning off trimming unconditionally, and should fix up our prompts to no longer have trailing space

i have a WIP branch where i've started to do this. so, assuming i get it working, i can merge into main before this and you can update your change here to just set `trim: false`.

remember to also set it inside the `restoreCursor` method though, too

43081j:
#351

btea:
Thanks for your review and work. Ok, let's close this PR.

43081j:
i actually meant we should still use this PR once i merge the other one

but no worries, i can make the change in my PR too 👍

btea:
It seems I misunderstood what you meant, but you can put this modification together with your PR. 👍

PR #351, opened 2025-06-22
fix: remove all trailing space from output

43081j:
Various prompts output trailing space in some situations.

This removes all such space and fixes a couple of bugs that were discovered while doing so.

Bugs which were fixed:

- Autocomplete had an empty line between the title and the rest. Now this is a bar
- Multi-select validation errors were not aligned after the first line
- Autocomplete `isNavigating` was not reset when pressing a non-navigation key in non-multiple mode


dreyfus92 (approved):
LGTM ✅

dreyfus92:
Just fix the tests 😁

PR #352, opened 2025-07-07
feat: add number prompt, fix note component, change directory structure, add feature to multiselect prompt

hyperz111:
# feat: add number prompt, fix note component, change directory structure, add feature to multiselect prompt

> [!NOTE]
> I **CANNOT** install `biome`

## What's Changes?

1. Add number prompt to `@clack/prompts`.
2. Change directory structure of `@clack/prompts` (opinionated).
3. Fix `note` component overflow glitch.
4. Add **toggle invert** for `multiselect` prompt by press `i` key.
5. Add **toggle all & invert** for `group-multiselect` prompt.
6. Add `sisteransi` and `picocolors` as `devDependencies`, it's mean `bundling` (it's works).
7. Add `wrap-ansi`.
8. And other....


43081j:
individually, most of these changes are probably fine, but they need splitting up into separate changes, separate PRs.

repo-level changes, like the structure, also need to be agreed by the team. so would be better as an issue rather than a PR for now

some of what you've changed are bug fixes and should be fixed. others are features we need discussion around to decide if we even want such a feature

if you open individual PRs, we can discuss in the PRs

hyperz111:
> individually, most of these changes are probably fine, but they need splitting up into separate changes, separate PRs.
> 
> repo-level changes, like the structure, also need to be agreed by the team. so would be better as an issue rather than a PR for now
> 
> some of what you've changed are bug fixes and should be fixed. others are features we need discussion around to decide if we even want such a feature
> 
> if you open individual PRs, we can discuss in the PRs

Hmm ok thanks

PR #354, opened 2025-07-10
feat(autocomplete-multiselect): allow select all and selecting inverse using shift up/down

tefkah:
This PR adds the ability to select all/inverse items for the `autocompleMultiselect` prompt, using Shift+Up (select all) and Shift+Down (select inverse)

See this (badly) in action here
https://github.com/user-attachments/assets/1a1275c7-51a2-4017-9232-8e8b09b748f5

## Caveats/things that can be changed

- The select all/select inverse behavior only applies to the filtered down list, not the whole list. I thought this would be more predictable.
- This behavior could be an option, but i decided to just do it by default
- When selecting all when everything is selected, everything will be deselected again
- Shift+Up and down were used bc these were the easiest to implement without touching the `prompt` base class. 
	- Initially i wanted to do `ctrl+a` and `ctrl+i`, but `ctrl+i` = `tab` in node readline. 
	- Also relying on the users keyboard layout like that did not seem like a good idea. 
	- I also thought about `[` and `]`, but could not erase those.
	  - there is a check in `Prompt` that erases `_actionKey`s, but that relies on the key having a name, which `[` and `]` do not have
	  - i could do it manually by also writing `ctrl+h` to `Prompt.rl`, but `Prompt.rl` is private.
  - I also tried Left and Right, but could not move the cursor back to the correct spot without access to `Prompt.rl`

Let me know what you think and whether i should change anything!

### DIFF
--- a/.changeset/happy-spoons-push.md
+++ b/.changeset/happy-spoons-push.md
@@ -1,0 +1,6 @@
+---
+"@clack/prompts": minor
+"@clack/core": minor
+---
+
+add inversion and selecting all options for autocomplete multiselect
--- a/packages/core/src/prompts/autocomplete.ts
+++ b/packages/core/src/prompts/autocomplete.ts
@@ -140,6 +140,8 @@ export default class AutocompletePrompt<T extends OptionLike> extends Prompt<
 		const isUpKey = key.name === 'up';
 		const isDownKey = key.name === 'down';
 		const isReturnKey = key.name === 'return';
+		const isLeft = key.name === 'left';
+		const isRight = key.name === 'right';
 
 		// Start navigation mode with up/down arrows
 		if (isUpKey || isDownKey) {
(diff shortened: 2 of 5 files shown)


dreyfus92 (commented):
This looks good to me, what do you think? @43081j 👀

43081j:
this is a difficult one, the bindings dont seem intuitive to me (shift+up being select all).

but we can't have ctrl-a since it is likely to be bound outside of what's running in the terminal (i.e. the terminal app probably will catch that, or tmux in my case).

additionally, you need to be able to type into this prompt, so it can't be `shift+{any readable character}`.

i wonder if we could have `shift-A` in navigation mode, i.e. if you've pressed tab or up/down. then you press shift-a, you select all

dreyfus92:
also cc. @natemoo-re some thoughts whenever you have a chance 🫣

tefkah:
@43081j yeah i was not very happy with these specific bindings either, but almost i encountered some limitation with most others (`Ctrl+a` like you mentioned, but especially `Ctrl-i` as that is bound to Backspace in `node` already`).

Only allowing select all/invert during navigation is an interesting idea, but i do think that would be a worse experience than having a globally accessible binding, and also kind of difficult to communicate (should we only show the hint during navigation then?).

I was also thinking of maybe `[`/`]` (or `{`/`}`), but that also has obvious problems of possibly conflicting with the search text.

Any keybind that also maps onto a searchable character would likely also need to erase that character immediately after, which I couldn't figure out how to do personally, but i'm sure someone here does!

Just to throw out some further suggestions for select all/invert respectively
- <code>\`</code>/`~`
- `Ctrl-[`/`Ctrl-]`
- `<`/`>` (hard to hint, looks like arrow keys)
- `\`/`|`

thank you all for taking the time to look at this~

PR #358, opened 2025-07-19
feat: add invert selection for multiselect prompt

hyperz111:
Create invert selection in `multiselect` by pressing `i` key like `inquirer` and `enquirer`.


tefkah:
great idea, would work well with #354 . might be worth having the same input for both, altho `i` wouldn't work for `autocomplete`

hyperz111:
> great idea, would work well with #354 . might be worth having the same input for both, altho `i` wouldn't work for `autocomplete`

Oh thanks

hyperz111:
@natemoo-re, please ping 🙏

hyperz111:
Hello @43081j, can you merge this?

43081j:
sorry just catching up on this

can you please add a test? in the test, we should emit the `i` key and ensure the selection is what we expect

hyperz111:
> sorry just catching up on this
> 
> can you please add a test? in the test, we should emit the `i` key and ensure the selection is what we expect

Oh ok 👍

PR #359, opened 2025-07-24
fix: wrap spinner output

### LINKED ISSUE #237: [Bug] Spinner message overflowing to the next line causes duplication
**Environment**

- OS: Windows
- Node Version: v18.20.4
- Package: `@clack/prompts`
- Package Version: 0.8.2

**Describe the bug**
If the length of a message displayed using spinner (start) is as long as the terminal window, where the loading dots i.e. `...` overflow to the next line in the terminal, it causes message to be duplicated while the spinner is running.

**To Reproduce**

Image

- Use spinner.start() with a message as long as the terminal window.
- See the overflow.

**Expected behavior**
Should overflow to the next line without any duplication.

43081j:
Wraps the spinner output and erases the previous frame based on the
wrapped value.

Fixes #237


dreyfus92 (approved):
LGTM 🫡

PR #363, opened 2025-08-03
feat: box prompt

43081j:
Adds a new `box` prompt which works like so:

```ts
prompts.box(
  'This is a message',
  'This is a title',
  {
    rounded: true
  }
);
```

## For reviewers

Feedback i'd like on this:

- Does `includePrefix` make sense? Basically its to tell the box to output the usual `|` bar before it, so if it is part of a group of prompts it will fit in. If you set it to `false` (the default), it won't output this
- This is very similar to the `note` prompt. It may be argued that we should just add the functionality to that instead of having a separate prompt

### DIFF
--- a/.changeset/shy-ideas-shout.md
+++ b/.changeset/shy-ideas-shout.md
@@ -1,0 +1,5 @@
+---
+"@clack/prompts": minor
+---
+
+Added new `box` prompt for rendering boxed text, similar a note.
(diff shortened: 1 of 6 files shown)


43081j:
i think you're right but i want to avoid some kind of `borderColor` setting

maybe we should do what `note` does and provide a `format` option, and a `formatBorder` option or some such thing

basically so you bring your own formatting, like `formatBorder: (symbol) => styleText('blue', symbol)`

43081j:
i've added a `formatBorder` option, so you can do this:

```ts
box('some message', 'some title', {
  formatBorder: styleText.bind(null, 'red'), // or picocolors.red
});
```

dreyfus92 (commented):
let's ship it 🤘🏻

PR #364, opened 2025-08-05
feat: add param to clear input after password prompt error

kajoseph:
This PR adds a `clearOnError` option to the password prompt to enforce typing the whole password again on an error. This is particularly useful when `mask: ''` (i.e. a hidden input) is given.

e.g. Consider this case
```
const pw = await prompt.password({
  message: 'Please enter your password:',
  mask: '',
  validate: (val) => val === 'test123' ? undefined : 'Password must be "test123"'
});
```

If the user enters "password" and hits return, an (expected) error `Password must be "test123"` will appear, like so:
image
The user then enters "test123" and hits return. They will get the same error because the value is actually "passwordtest123" which is not apparent with a hidden input.

This PR adds a `clearOnError` boolean option to have the input clear out after an error to avoid the confusion above.

### DIFF
--- a/.changeset/solid-goats-shop.md
+++ b/.changeset/solid-goats-shop.md
@@ -3,4 +3,4 @@
 "@clack/core": minor
 ---
 
-Add input reset to password prompt
+Add `clearOnError` option to password prompt to automatically clear input when validation fails


packages/core/src/prompts/prompt.ts
```diff
@@ -186,6 +186,14 @@ export default class Prompt<TValue> {
 		}
 	}
 
+	protected _resetUserInput(): void {
+		this.rl?.write(null, { ctrl: true, name: 'u' });
+		if (this._track) {
```
43081j, line 191: can we just call `this._setUserInput('')` here?

packages/prompts/src/password.ts
```diff
@@ -6,6 +6,7 @@ export interface PasswordOptions extends CommonOptions {
 	message: string;
 	mask?: string;
 	validate?: (value: string | undefined) => string | Error | undefined;
+	reset?: boolean;
```
43081j, line 9: i wonder if we should be clearer about this - naming it something like `resetOnError` or `clearOnError`

43081j:
im thinking it may make more sense to call it `clear` rather than `reset`, since it does empty the input.

if a prompt has a default value, i'd expect `reset` to reset back to that. while i'd expect `clear` to empty the input

.changeset/solid-goats-shop.md
```diff
@@ -0,0 +1,6 @@
+---
+"@clack/prompts": minor
+"@clack/core": minor
+---
+
+Add input reset to password prompt
```
dreyfus92, line 6:
```suggestion
Add `clearOnError` option to password prompt to automatically clear input when validation fails
```

dreyfus92 (commented):
Awesome work @kajoseph, thank you for taking the time on working on this. I left a small suggestion to make the changeset a bit clearer and concise.

43081j (approved):

PR #385, opened 2025-09-03
feat: support disabled for select and multiselect prompt

### LINKED ISSUE #184: [Request] `Select` or `Multi-Select` provide `disable` option in `@clack/prompts`
As the title suggests, is it possible to provide a `disable` props in the options so that we can skip selecting it when we select it.

```diff
{
    value: 'value',
    label: 'label',
+   disabel: true
}
```

I want to show it when disabled without filtering it out.

chouchouji:
close #184


chouchouji:
@43081j Hi, James. I have submitted this pr to support #184, Please review it when you have time. Thanks!
 Test cases will be added after we have finalized the features to be implemented.

43081j:
i've had a quick read through but ill have to come back to it (busy day!)

it seems good but im contemplating whether the core needs to know about disabled or not - since its a rendering thing rather than an underlying logic thing. though we do need to know for the cursor changes

anyway ill take a deeper look soon and figure it out 👍

chouchouji:
> i've had a quick read through but ill have to come back to it (busy day!)
> 
> it seems good but im contemplating whether the core needs to know about disabled or not - since its a rendering thing rather than an underlying logic thing. though we do need to know for the cursor changes
> 
> anyway ill take a deeper look soon and figure it out 👍

OK, Thanks. If you have some good ideas, just add them in this pr.

PR #383, opened 2025-08-28
fix(note): hard wrap

hyperz111:
This PR contains a change i forgot add on #376. The change is **"Do hard wrap"**.

PR #386, opened 2025-09-03
fix: add missing null checks around values

43081j:
Also uses `getColumns` for note wrapping and changes the `types` script
to run `tsc` so we can catch these errors in CI in future.


hyperz111:
Hello @43081j, can you check #383? (If you have a time)

PR #387, opened 2025-09-04
refactor(progress): remove unnecessary return statement in start func…

chouchouji:
…tion

The return value of `spinner.start` is `undefined`. I think getting this return value is not useful for developers. So can we remove `return` safely ?


43081j (approved):
makes sense to me 👍

chouchouji:
> makes sense to me 👍
> 
> can you add a changeset?

Yeah, I have added a changeset. Thanks for your review !

43081j (approved):

PR #389, opened 2025-09-05
docs: add type assert for text validation example

chouchouji:
The type of `value` should be `string`,  maybe we can add type assert to pass this type check.

image

### DIFF
--- a/examples/basic/text-validation.ts
+++ b/examples/basic/text-validation.ts
@@ -9,7 +9,7 @@ async function main() {
 		message: 'Enter your name (letters and spaces only)',
 		initialValue: 'John123', // Invalid initial value with numbers
 		validate: (value) => {
-			if (!/^[a-zA-Z\s]+$/.test(value)) return 'Name can only contain letters and spaces';
+			if (!value || !/^[a-zA-Z\s]+$/.test(value)) return 'Name can only contain letters and spaces';
 			return undefined;
 		},
 	});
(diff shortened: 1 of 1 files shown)


43081j:
the `value` can be `undefined` because you can press `<return>` before typing anything

so it is possible that `validate(undefined)` gets called

we should be doing something like:

```ts
if (!value || !value.test(regex)) {
  // error
}
```

chouchouji:
> the `value` can be `undefined` because you can press `<return>` before typing anything
> 
> so it is possible that `validate(undefined)` gets called
> 
> we should be doing something like:
> 
> ```ts
> if (!value || !value.test(regex)) {
>   // error
> }
> ```

Thanks for your reply. The example has an `initialValue` value. I try to press it.  It is not `undefined`.  If I have some wrong steps, please tell me.

钉钉录屏_2025-09-05 173032

43081j:
it does have an initial value so it means in this particular case, it can't be `undefined`

but as far as typescript is concerned, it can still be `undefined` because you might not have set the initial value

the types are currently right even though you've managed to make runtime code that never sets it as `undefined`

chouchouji:
> it does have an initial value so it means in this particular case, it can't be `undefined`
> 
> but as far as typescript is concerned, it can still be `undefined` because you might not have set the initial value
> 
> the types are currently right even though you've managed to make runtime code that never sets it as `undefined`

I get it. Initially, I added a non-null check. But after testing, I am confused that it should be `string` type. Thanks for your reply again. 🌹

PR #384, opened 2025-08-29
feat(autocomplete): add wrapping and window limits

43081j:
Adds two things to the autocomplete prompt:

1. Wrapping (in that options can now wrap across multiple lines)
2. Automatically shrinking the max items displayed if they don't fit on the screen

The limit options helper has become quite a bit more complicated to handle the latter :grimacing: but that's _probably_ ok.

## TODO

- add tests
- ???

PR #393, opened 2025-09-08
feat: support disabled for select and multiselect prompt 

### LINKED ISSUE #184: [Request] `Select` or `Multi-Select` provide `disable` option in `@clack/prompts`

chouchouji:
close #184


43081j:
I haven't forgotten this FYI, i'm just waiting until I land #384 before i re-review

43081j:
i pushed a few changes @chouchouji 

im not too sure about how best to render this though. the black circles don't look great. it also feels odd skipping over options

i wonder if we should allow you to focus a disabled option but not select it. ill bring it up in discord and see what people think

43081j (approved):

chouchouji:
> i pushed a few changes @chouchouji
> 
> im not too sure about how best to render this though. the black circles don't look great. it also feels odd skipping over options
> 
> i wonder if we should allow you to focus a disabled option but not select it. ill bring it up in discord and see what people think

Thanks for your hard work.

PR #395, opened 2025-09-09
feat: add `clearPromptOnDone` for some prompts

hyperz111:
In this PR, i add `clearPromptOnDone` option for some prompts like on `inquirer`.
The prompt who get `clearPromptOnDone` option are:

- `autocomplete`
- `autocompleteMultiselect`
- `confirm`
- `groupMultiselect`
- `multiselect`
- `password`
- `path`
- `selectKey`
- `select`
- `text`

### DIFF
--- a/packages/prompts/src/autocomplete.ts
+++ b/packages/prompts/src/autocomplete.ts
@@ -1,8 +1,9 @@
 import { AutocompletePrompt } from '@clack/core';
 import color from 'picocolors';
-import { cursor } from "sisteransi";
+import { cursor } from 'sisteransi';
 import {
 	type CommonPromptOptions,
+	clearPrompt,
 	S_BAR,
 	S_BAR_END,
 	S_CHECKBOX_INACTIVE,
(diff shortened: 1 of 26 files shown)


hyperz111:
@43081j, do we need this feature?

43081j:
sorry i have taken a while to get to this!

honestly we probably don't need it yet unless you've seen a use case somewhere. i want to avoid adding functionality unless there is demand for it as thats how alternative libraries ended up bloated

if we do one day need this feature, i'd be tempted to just document that you can do it yourself at the call site:

```ts
const input = await text(opts);
clearTerminal(); // pull this from sister ansi or something
```

hyperz111:
> sorry i have taken a while to get to this!
> 
> honestly we probably don't need it yet unless you've seen a use case somewhere. i want to avoid adding functionality unless there is demand for it as thats how alternative libraries ended up bloated
> 
> if we do one day need this feature, i'd be tempted to just document that you can do it yourself at the call site:
> 

So, i should close this?

43081j:
I'll close it for now and we can easily pick it back up one day if needed 👍

The effort is much appreciated still

PR #402, opened 2025-10-03
feat(spinner): nothing appear when stop message is undefined

### LINKED ISSUE #401: [Request] Empty message `spinner.stop()` should print nothing
Currently, when using `spinner.stop()` without any message, an empty line with the success symbol is printed out. The way I would like to use it is; I have a command that can have three outcomes: pass, soft fail and hard fail. Hard fail uses `spinner.stop(msg, 1)`, but during soft fail I would like to have an option to not print out anything on spinner stop and use something like `log.warn()` instead.

I would like for the command to not print out the empty line with just the success symbol when the message is empty, or a way to toggle this, if possible.

I've considered not using spinner for this and just using `log.message()` instead, but I like the "loading" element that spinner provides.

**Additional context**
\-

hyperz111:
Fixes #401


43081j:
my concern here is the signature doesn't align anymore:

```ts
const stop = (msg = '', code = 0, clear = false): void
```

`msg` is basically ignored if `clear === true`. i.e. `stop("foo bar", 0, true)` wouldn't use `"foo bar"` at all

i wonder if we should have a `clear` method instead, for this reason. and it would be implied clearing the spinner also means stopping it

thoughts @dreyfus92 ?

hyperz111:
> my concern here is the signature doesn't align anymore:
> 
> ```ts
> const stop = (msg = '', code = 0, clear = false): void
> ```
> 
> `msg` is basically ignored if `clear === true`. i.e. `stop("foo bar", 0, true)` wouldn't use `"foo bar"` at all
> 
> i wonder if we should have a `clear` method instead, for this reason. and it would be implied clearing the spinner also means stopping it
> 
> thoughts @dreyfus92 ?

If we follow the #401 suggestion, this will break `progress` component.

dreyfus92:
hey peeps, i agree with @43081j. it would be best to create a `clear()` method which doesn't create confusion about params behavior, and it is also backward compatible (doesn't change existing `stop()` signature). 

```ts
export interface SpinnerResult {
    start(msg?: string): void;
    stop(msg?: string, code?: number): void;
    clear(code?: boolean): void; // NEW: silently stops spinner
    message(msg?: string): void;
    readonly isCancelled: boolean;
}
```

43081j:
> this will break `progress` component

can you explain why it would break the progress prompt?

could we not just add `clear` there too?

hyperz111:
> can you explain why it would break the progress prompt?

If we change the `stop` method.

**BEFORE:**

```ts
const stop = (msg = '', code = 0): void => {
	if (!isSpinnerActive) return;
	isSpinnerActive = false;
	clearInterval(loop);
	clearPrevMessage();
	const step =
		code === 0
			? color.green(S_STEP_SUBMIT)
			: code === 1
				? color.red(S_STEP_CANCEL)
				: color.red(S_STEP_ERROR);
	_message = msg ?? _message;
	if (indicator === 'timer') {
		output.write(`${step}  ${_message} ${formatTimer(_origin)}\n`);
	} else {
		output.write(`${step}  ${_message}\n`);
	}
	clearHooks();
	unblock();
};
```

**AFTER:**

```diff
- const stop = (msg = '', code = 0): void => {
+ const stop = (msg, code = 0): void => {
	if (!isSpinnerActive) return;
	isSpinnerActive = false;
	clearInterval(loop);
	clearPrevMessage();
	const step =
		code === 0
			? color.green(S_STEP_SUBMIT)
			: code === 1
				? color.red(S_STEP_CANCEL)
				: color.red(S_STEP_ERROR);
-	_message = msg ?? _message;
-    if (indicator === 'timer') {
-		output.write(`${step}  ${_message} ${formatTimer(_origin)}\n`);
-	} else {
-		output.write(`${step}  ${_message}\n`);
+	if (msg != null) {
+		_message = msg;
+		if (indicator === 'timer') {
+			output.write(`${step}  ${_message} ${formatTimer(_origin)}\n`);
+		} else {
+			output.write(`${step}  ${_message}\n`);
+		}
	}
	clearHooks();
	unblock();
};
```

When we run the test (maybe we must update the test):

```console
  Snapshots  28 failed
 Test Files  2 failed | 12 passed (14)
      Tests  28 failed | 356 passed (384)
   Start at  13:18:24
   Duration  15.13s (transform 5.40s, setup 0ms, collect 29.24s, tests 10.29s, environment 48ms, prepare 19.68s)
```

> could we not just add `clear` there too?

In some spinner libraries, `clear` method is **JUST** for cleaning the spinner and **NOT** stopping the spinner.
If we don't want to follow some libraries convention, maybe we can add `clear` method instead.

43081j:
@delucis has also opened #405 

i wonder if we should try land that first, then add a `clear` method too

the one other popular spinner (ignoring `ora` since it is going away) - `nanospinner`, doesn't have a `clear` method. none of the ones i know have this capability.

so i'd be happy to have a `clear` method for this

PR #405, opened 2025-10-20
Refactor the API for stopping spinners & progress bars

delucis:
As [discussed on Discord](https://discord.com/channels/1161404915181162619/1289689029637443604/1429966860258119791), this PR proposes a change to the API for stopping spinners and progress bars.

Currently these use a slightly awkward API using a positional argument to pass a numeric code to set whether the stop is successful or the result of an error or cancellation, e.g. `spinner.stop(undefined, 1)`.

This PR proposes removing the error code form in favour of distinct `spinner.cancel()` and `spinner.error()` methods.

This is a breaking change for anyone using the code API currently.

### Notes 

This PR does not change any of the existing semantics around how “cancel” and “error” are represented and preserves the current behaviour where a “cancel” code shows a red square and an “error” code shows a yellow triangle.

This is different from the `.error()` method of the `log` utility, which shows a red square.

An alternative refactor could be to rename to `.stopAndWarn()` (current “error” behaviour) and `.stopAndError()` (current “cancel” behaviour), or something like that, which would align with how `log.warn` and `log.error` look.

Or, another alternative would be to switch to an object argument like `.stop({ type: 'warn' })`, which could also resolve the awkwardness of the current positional style and avoid multiple methods.

### Docs

This will require an update to https://bomb.sh/docs/clack/packages/prompts/#spinner

### DIFF
--- a/packages/prompts/test/__snapshots__/progress-bar.test.ts.snap
+++ b/packages/prompts/test/__snapshots__/progress-bar.test.ts.snap
@@ -167,6 +167,34 @@ exports[`prompts - progress (isCI = false) > stop > renders message 1`] = `
 ]
 `;
 
+exports[`prompts - progress (isCI = false) > stop > renders message when cancelling 1`] = `
+[
+  "<cursor.hide>",
+  "[90m│[39m
+",
+  "[35m◒[39m  [35m[39m[2m━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━[22m ",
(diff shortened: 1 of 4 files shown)


dreyfus92 (approved):
this is very solid, thank you Chris, welcome to the clack gang! 🤘🏻

43081j:
looks good to me 👍 

can you add a test or two for passing a non-empty message?

delucis:
> can you add a test or two for passing a non-empty message?

Done!

PR #409, opened 2025-11-06
feat: add withGuide option

43081j:
This adds a new `withGuide` option which decides if the standard clack left border is rendered or not.

Not all prompts implement this yet. We can add support for more prompts over time.

### DIFF
--- a/packages/prompts/test/__snapshots__/text.test.ts.snap
+++ b/packages/prompts/test/__snapshots__/text.test.ts.snap
@@ -79,7 +79,6 @@ exports[`text (isCI = false) > global withGuide: false removes guide 1`] = `
 
 ",
   "<cursor.backward count=999><cursor.up count=3>",
-  "",
   "<erase.down>",
   "[32m◇[39m  foo
 ",
@@ -289,7 +288,6 @@ exports[`text (isCI = false) > withGuide: false removes guide 1`] = `
 
 ",
   "<cursor.backward count=999><cursor.up count=3>",
-  "",
   "<erase.down>",
   "[32m◇[39m  foo
 ",
@@ -378,7 +376,6 @@ exports[`text (isCI = true) > global withGuide: false removes guide 1`] = `
 
 ",
   "<cursor.backward count=999><cursor.up count=3>",
-  "",
   "<erase.down>",
   "[32m◇[39m  foo
 ",
@@ -588,7 +585,6 @@ exports[`text (isCI = true) > withGuide: false removes guide 1`] = `
 
 ",
   "<cursor.backward count=999><cursor.up count=3>",
-  "",
   "<erase.down>",
   "[32m◇[39m  foo
 ",


dreyfus92 (approved):
LGTM 😁

43081j:
the global settings have their downsides too, since you may not always want to _every_ prompt to follow these rules

maybe in a follow up change, we can make the global settings inherit some of the shared settings, and the inline option takes precedence. i.e. if you set global `withBorder` but pass `withBorder: true`, it'll override it

packages/prompts/src/box.ts
```diff
@@ -68,7 +67,7 @@ export const box = (message = '', title = '', opts?: BoxOptions) => {
 	const titlePadding = opts?.titlePadding ?? 1;
 	const contentPadding = opts?.contentPadding ?? 2;
 	const width = opts?.width === undefined || opts.width === 'auto' ? 1 : Math.min(1, opts.width);
-	const linePrefix = opts?.includePrefix ? `${S_BAR} ` : '';
+	const linePrefix = opts?.withBorder === false ? '' : `${S_BAR} `;
```
delucis: Does this means boxes would render without a left border? i.e. something like

```
╭───────────────────────╮
  example box           │
  some more text...     │
╰───────────────────────╯
```

I’m not sure, but this one feels like even when rendering without a left border, you’d still want boxes to render with the four sides intact. At least if this is a generic option set at a global level. (And if a local thing like `box('example', null, { withBorder: false })`, then I’d expect _no_ border at all, not only the left border to be missing.

Maybe there’s an argument that what we’re describing as a “border” is not really a _border_ in the generic rendering sense. It has a UI function in connecting prompts, guiding the eye from step to step, etc. So something like `withGuide` could be a more appropriate naming? And then that would make it more obvious that a box would not be impacted (but might have other styling hooks given it genuinely is a _bordered_ element).

packages/prompts/src/text.ts
```diff
@@ -20,7 +20,9 @@ export const text = (opts: TextOptions) => {
 		signal: opts.signal,
 		input: opts.input,
 		render() {
-			const title = `${color.gray(S_BAR)}\n${symbol(this.state)}  ${opts.message}\n`;
+			const withBorder = opts.withBorder !== false;
+			const titlePrefix = withBorder ? `${color.gray(S_BAR)}\n${symbol(this.state)}  ` : '';
```
delucis: IIUC this means no symbols prefixing prompts when disabling the border? It’s personal preference I guess, but I’d still expect some kind of leading marker even when disabling borders/guides. These markers can be helpful indicators of state.

For example, to mirror something a bit like what `prompts` does, state is indicated via something like
```
? What is your name?
  Placeholder
```
And then post-confirm
```
✔︎ What is your name?
  Chris
```

These could definitely just be the standard clack symbols by default but the left border/guide seems a distinct UI element from these status indicators to me.

delucis (commented):
Finally got a minute to take a look! Thanks @43081j!

43081j:
i've renamed this to `withGuide` and have also reintroduced the prefix to text prompts

@delucis could you take another look?

43081j:
ok @natemoo-re i've added a global `withGuide` setting now too

i'll aim to merge this today so let me know if any of you have any feedback

PR #415, opened 2025-11-16
feat: add theme option

### LINKED ISSUE #345: [Request] add a style representation to all prompts
It could be useful to expose the styling (symbols and colours) of each prompt without having to expose actual rendering.

Something like this may be enough:

```ts
await text({
  message: 'whats up?',
  style: {
    label: picocolors.red,
    someSymbol: () => '#'
  }
});
```

although a super-variable object is making my deopt senses go off. but it is more useable than `style: [{name: 'label', fn: picocolors.red}]`

hyperz111:
In this PR, i add `theme` option to some prompts:

- `autocomplete`
- `autocomplete-multiselect`
- `confirm`
- `group-multi-select`
- `multiselect`
- `password`
- `select-key` (NOT perfect)
- `select`
- `text`

~See `ThemeOptions` in `common.ts`.~

fixes #345

### DIFF
--- a/packages/prompts/src/autocomplete.ts
+++ b/packages/prompts/src/autocomplete.ts
@@ -1,6 +1,15 @@
 import { AutocompletePrompt } from '@clack/core';
 import color from 'picocolors';
-import { type CommonOptions, type RadioTheme, type CheckboxTheme, getThemeColor, getThemePrefix, extendStyle, S_BAR, S_BAR_END } from './common.js';
+import {
+	type CheckboxTheme,
+	type CommonOptions,
+	extendStyle,
+	getThemeColor,
+	getThemePrefix,
+	type RadioTheme,
+	S_BAR,
+	S_BAR_END,
+} from './common.js';
 import { limitOptions } from './limit-options.js';
 import type { Option } from './select.js';
 
@@ -53,7 +62,7 @@ type AutocompleteSharedOptions<Value, TStyle> = CommonOptions<TStyle> & {
 	 * Validates the value
 	 */
 	validate?: (value: Value | Value[] | undefined) => string | Error | undefined;
-}
+};
 
 export interface AutocompleteOptions<Value> extends AutocompleteSharedOptions<Value, RadioTheme> {
 	/**
@@ -193,7 +202,8 @@ export const autocomplete = <Value>(opts: AutocompleteOptions<Value>) => {
 };
 
 // Type definition for the autocompleteMultiselect component
-export interface AutocompleteMultiSelectOptions<Value> extends AutocompleteSharedOptions<Value, CheckboxTheme> {
+export interface AutocompleteMultiSelectOptions<Value>
+	extends AutocompleteSharedOptions<Value, CheckboxTheme> {
 	/**
 	 * The initial selected values
 	 */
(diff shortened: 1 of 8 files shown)


43081j:
the approach isn't far off, but i don't think it makes sense to have a single set of options

each prompt has its own style. they only cross over on some things, such as the left guide/border.

i'd expect `CommonOptions` to have a type parameter for the style, or multiple inheritance of interfaces:

```ts
interface CommonOptions<TStyle> {
  // ...
  style: TStyle;
}

// or

interface Whatever extends CommonOptions, StyleOptions<WhateverStyle> {
  // ..
}
```

43081j:
you're not far off but its not quite what i meant

here:

```ts
interface TextPromptStyle {
  someTextPromptSpecificStyle: string;
  anotherOne: string;
}

interface TextPromptOptions extends CommonOptions<TextPromptStyle> {
  // ..
}
```

every prompt would have its own style interface. if it doesn't have any of its own, it would use a `BaseStyle` or some such thing

the style will _always_ be a flat object. the only reason you had to introduce deep merge logic is because you chose to make the shape nested. just don't bother - keep it flat and you can throw the deep merge stuff away.

`CommonOptions` should basically be `CommonOptions<TStyle extends Record<string, string>>`

hyperz111:
Like this?

```ts
// Before 
interface CheckboxTheme {
	selected?: {
		active?: string;
		inactive?: string;
	};
	unselected?: {
		active?: string;
		inactive?: string;
	};
	disabled?: string;
}

// After
interface CheckboxThemeFlat {
	selectedActive?: string;
	selectedInactive?: string;
	unselectedActive?: string;
	unselectedInactive?: string;
	disabled?: string;
}
```

43081j:
yes exactly, but there'd be no such thing as a "checkbox theme"

there would be a text prompt theme, a select prompt theme, etc.

if any of those share some properties, just have an internal interface they both extend

dreyfus92 (commented):
hey @hyperz111, amazing work on this! however, I would suggest a different approach given the scope of changes needed.

what @43081j is suggesting is a significant architectural refactor, the nested structure needs to be completely flattened. trying to tackle all 8 prompts at once makes this PR very difficult to review.

break this into separate PRs, starting with just the autocomplete prompt. once we get that pattern right and merged, the others will follow the same approach and be much easier to review.

for this PR, I'd suggest:

- focus only on `autocomplete.ts` and `autocompleteMultiselect.ts`
- implement the flat theme structure James outlined (no nesting, prompt-specific styles)
- remove `@fastify/deepmerge` and use simple shallow merging like object spreading
- get this pattern approved and merged first

then follow up with separate PRs for:

- `select.ts` and `multiselect.ts`
- `confirm.ts`
- `password.ts` and `text.ts`
- `group-multiselect.ts`
- `select-key.ts`

43081j:
having another think about this. i wonder if we should have something like this:

```ts
interface GlobalTheme {
  formatGuide?: (str: string) => string;
}

interface ThemeOptions<T> {
  theme?: T & GlobalTheme;
}

interface CommonOptions {
  // existing common options interface
}

// ...

export interface TextTheme {
 formatCancel?: (str: string) => string;
}
  
export interface TextOptions extends CommonOptions, ThemeOptions<TextTheme> {
  message: string;
  placeholder?: string;
  defaultValue?: string;
  initialValue?: string;
  validate?: (value: string | undefined) => string | Error | undefined;
}
```

so you'd use it like this:

```ts
text({
  theme: {
    formatGuide: (str) => colors.red(str), // make the left guide line/border red
  }
});
```

the names need some bikeshedding but this pattern is what i had in mind. and fully agree we need to really strip this down to granular changes

hyperz111:
> break this into separate PRs, starting with just the autocomplete prompt. once we get that pattern right and merged, the others will follow the same approach and be much easier to review.

I don't want to do that, but if it's for making the review easy. Maybe i can

hyperz111:
I close this for now

43081j:
i think a good place to pick this back up would be to apply styling only to one prompt, e.g. the text prompt

as part of that, we will have to implement the base theme options and what not.

its a pretty big feature overall so might take a fair bit of discussion, back/forth, etc to get it right

hyperz111:
> i think a good place to pick this back up would be to apply styling only to one prompt, e.g. the text prompt
> 
> as part of that, we will have to implement the base theme options and what not.
> 
> its a pretty big feature overall so might take a fair bit of discussion, back/forth, etc to get it right

I maybe make a new in this weekend

PR #426, opened 2025-12-04
feat(@clack/prompts): add theme support for text prompt customization

dreyfus92:
Add theme support for the text prompt. Users can now customize the colors of symbols, guide lines, and error messages by passing a `theme` option.


43081j:
This generally looks pretty much how I was hoping we would implement it 😀

I want to try it locally before we merge though as I'm curious if we handle the fact these symbols and what not are variable length now

Can you manually do some testing? We don't need new test cases, just manually try it out with smaller terminals etc. To ensure the wrapping still works

packages/prompts/src/common.ts
```diff
@@ -73,3 +73,38 @@ export interface CommonOptions {
...
+
+export type ColorFormatter = (str: string) => string;
+
+/**
+ * Global theme options shared across all prompts.
+ * These control the common visual elements like the guide line.
+ */
+export interface GlobalTheme {
+	/** Format the left guide/border line (default: cyan) */
+	formatGuide?: ColorFormatter;
```
hyperz111: Maybe we can use `guide` instead?

dreyfus92:
closing since this became stale

PR #444, opened 2026-01-19
feat(prompts): add custom filter option to autocomplete

danielroe:
## Summary

a custom `filter` opt is support in multiselect - would be nice to do the same for the autocomplete version 🙏 

## Motivation

related PR: https://github.com/nuxt/cli/pull/1180

The `@clack/core` package already supports custom filter functions via the `AutocompletePrompt` class, but this wasn't exposed in the higher-level `@clack/prompts` API. This is useful for:

- Implementing fuzzy search (e.g., using `fzf` library)
- Custom matching logic (prefix-only, regex-based, etc.)
- Searching additional fields beyond label/hint/value

## Changes

- Added `filter?: (search: string, option: Option<Value>) => boolean` to `AutocompleteSharedOptions`
- Updated `autocomplete()` and `autocompleteMultiselect()` to pass through the custom filter, falling back to the default filter when not provided
- Added tests for custom filter functionality

## Example Usage

```typescript
import { autocompleteMultiselect } from '@clack/prompts'
import { Fzf } from 'fzf'

const options = [
  { value: 'apple', label: 'Apple' },
  { value: 'banana', label: 'Banana' },
]

const fzf = new Fzf(options, { selector: (opt) => opt.label })

const result = await autocompleteMultiselect({
  message: 'Select fruits',
  options,
  filter: (search, option) => {
    if (!search) return true
    return fzf.find(search).some((r) => r.item.value === option.value)
  },
})
```

### DIFF
--- a/.changeset/afraid-rabbits-grin.md
+++ b/.changeset/afraid-rabbits-grin.md
@@ -1,0 +1,5 @@
+---
+"@clack/prompts": minor
+---
+
+This adds a custom filter function to autocompleteMultiselect. It could be used, for example, to support fuzzy searching logic.


43081j (approved):
makes sense to me 👍 

can you add a changeset? (`npx changeset`)

danielroe:
done! ❤️

Issue #447, opened 2026-01-28
[Request] `date` prompt


davidwarrington:
**Is your feature request related to a problem? Please describe.**
When requesting date input, it would be nice to have a prompt other than text for it. For example, it _could_  be a text input that enforces a specific format, e.g. `YY/MM/DD`, where the user can also navigate between the editable regions using the left and right arrows, and use up and down arrows to increase/decrease the number in that region of the input. Alternatively there could be an actual datepicker UI for this.

I've been thinking about creating one for a side of desk project I have at work, but haven't been able to settle on what the UX ought to be for it. If this is something you'd be interested in having in the main library, I'd be happy to try building it myself but would appreciate some input on the UX side of things in particular.

**Describe alternatives you've considered**
Using the `validate` method on a text input. This can work fine, but I think a strict format that users can't escape from by typing whatever they like is better.

43081j:
sounds like a good idea to me 👍 

i think we should dig into it a bit to find out how feasible some kind of date picker UX is. but if it turns out to be a pain to render that nicely, we can go with the constrained input (only accepts dates and renders the `/` between the segments).

ill have a play around with some ideas too and let you know what seems possible

dreyfus92:
hey @davidwarrington i've got something working already, dunno if you had any time to work on this. I might publish in a bit a PR so you can take a look 😄

davidwarrington:
@dreyfus92 thanks for sharing. I've had very little time to look into this so glad somebody else has.

davidwarrington:
Closing since #448 was merged.

PR #448, opened 2026-01-31
feat(core, prompts): add DatePrompt for date input with customizable formats

dreyfus92:
Implements a date prompt that enforces a structured date format and supports the most common variants: `YYYY/MM/DD`, `MM/DD/YYYY`, and `DD/MM/YYYY`. 

related: #447

### DIFF
--- a/packages/core/src/prompts/date.ts
+++ b/packages/core/src/prompts/date.ts
@@ -272,7 +272,9 @@ export default class DatePrompt extends Prompt<Date> {
 		);
 
 		const newNum = isBlank
-			? (direction === 1 ? bounds.min : bounds.max)
+			? direction === 1
+				? bounds.min
+				: bounds.max
 			: clamp(bounds.min, num + direction, bounds.max);
 
 		const newSegmentValue = String(newNum).padStart(segment.len, '0');
@@ -323,9 +325,7 @@ export default class DatePrompt extends Prompt<Date> {
 
 			const firstBlank = segmentDisplay.indexOf('_');
 			const pos =
-				firstBlank >= 0
-					? firstBlank
-					: Math.min(this.#cursor.positionInSegment, segment.len - 1);
+				firstBlank >= 0 ? firstBlank : Math.min(this.#cursor.positionInSegment, segment.len - 1);
 			if (pos < 0 || pos >= segment.len) return;
 
 			const newSegmentVal = segmentDisplay.slice(0, pos) + char + segmentDisplay.slice(pos + 1);
--- a/packages/prompts/src/date.ts
+++ b/packages/prompts/src/date.ts
@@ -1,6 +1,6 @@
+import { styleText } from 'node:util';
 import type { DateFormatConfig, DateParts } from '@clack/core';
 import { DatePrompt, settings } from '@clack/core';
-import color from 'picocolors';
 import { type CommonOptions, S_BAR, S_BAR_END, symbol } from './common.js';
 
 export type DateFormat = 'YYYY/MM/DD' | 'MM/DD/YYYY' | 'DD/MM/YYYY';
@@ -35,17 +35,19 @@ function renderSegment(
 	if (isBlank) {
 		if (cursorInThis) {
 			for (let j = 0; j < label.length; j++) {
-				parts.push(j === cursor.positionInSegment ? color.inverse(' ') : color.dim(label[j]));
+				parts.push(
+					j === cursor.positionInSegment ? styleText('inverse', ' ') : styleText('dim', label[j])
+				);
 			}
 		} else {
-			parts.push(color.dim(label));
+			parts.push(styleText('dim', label));
 		}
 	} else {
 		for (let j = 0; j < value.length; j++) {
 			if (cursorInThis && j === cursor.positionInSegment) {
-				parts.push(value[j] === '_' ? color.inverse(' ') : color.inverse(value[j]));
+				parts.push(value[j] === '_' ? styleText('inverse', ' ') : styleText('inverse', value[j]));
 			} else {
-				parts.push(value[j] === '_' ? color.dim(' ') : value[j]);
+				parts.push(value[j] === '_' ? styleText('dim', ' ') : value[j]);
 			}
 		}
 	}
@@ -64,12 +66,11 @@ function renderDateFormat(
 		return config.format(parts);
 	}
 	const labels = config.segmentLabels ?? DEFAULT_SEGMENT_LABELS;
-	const sep = color.gray('/');
+	const sep = styleText('gray', '/');
 	const rendered = config.segments.map((seg, i) =>
 		renderSegment(parts[seg.type], i, cursor, labels[seg.type], state)
 	);
 	let result = rendered.join(sep);
-	// Add cursor block if beyond last segment
 	const lastSeg = config.segments[config.segments.length - 1];
 	if (
 		cursor.segmentIndex >= config.segments.length ||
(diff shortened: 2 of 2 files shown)


.changeset/tangy-mirrors-hug.md
```diff
@@ -0,0 +1,6 @@
+---
+"@clack/prompts": minor
+"@clack/core": minor
```
natemoo-re: Because we're v1.0, we're going to have to hold minor features until we're ready to do a release! Does a 2 week cadence seem fair?

Will spin up a discussion in Discord

packages/core/src/prompts/date.ts
```diff
@@ -0,0 +1,404 @@
...
+interface SegmentConfig {
+	type: 'year' | 'month' | 'day';
+	start: number;
+	len: number;
+}
+
+interface FormatConfig {
+	segments: SegmentConfig[];
+	displayTemplate: string;
+	format: (year: string, month: string, day: string) => string;
```
natemoo-re: Could we refactor this to be an object with `{ year, month, date }`? Object callbacks are more refactor-proof if we want to expose additional options in the future. 

General rule of thumb is that if you find yourself writing >=3 arguments, you should stick use an object instead.

packages/core/src/prompts/date.ts
```diff
@@ -0,0 +1,404 @@
...
+	'MM/DD/YYYY': ['month', 'day', 'year'],
+	'DD/MM/YYYY': ['day', 'month', 'year'],
+};
+
+// All formatters take (year, month, day) and return display string in format order
+const formatters: Record<DateFormat, (year: string, month: string, day: string) => string> = {
+	'YYYY/MM/DD': (y, m, d) => `${y}/${m}/${d}`,
+	'MM/DD/YYYY': (y, m, d) => `${m}/${d}/${y}`,
+	'DD/MM/YYYY': (y, m, d) => `${d}/${m}/${y}`,
+};
```
natemoo-re: This seems like a lot of duplication and segmentation logic. IMO preset for the display format are a `@clack/prompts` concern, but the core component should just accept a callback.

If we can't refactor it like that, can we store the actual `Date()` and leverage built-in formatters / segmenters with `Intl.DateTimeFormat`?

packages/core/src/prompts/date.ts
```diff
@@ -0,0 +1,404 @@
...
+		return 'There are only 12 months in a year';
+	}
+	if (seg.type === 'day') {
+		if (day < 1) return undefined; // incomplete
+		const daysInMonth = new Date(year || 2000, month || 1, 0).getDate();
+		if (day > daysInMonth) {
+			const monthName = month >= 1 && month <= 12 ? MONTH_NAMES[month - 1] : 'this month';
+			return `There are only ${daysInMonth} days in ${monthName}`;
+		}
+	}
```
natemoo-re: More strings for localization 😉

packages/core/src/prompts/date.ts
```diff
@@ -0,0 +1,404 @@
...
+	return undefined;
+}
+
+export interface DateOptions extends PromptOptions<string, DatePrompt> {
```
natemoo-re: The Date selector should operate on a `Date` rather than a string? This prompt should return a `Date` value. Might simplify some of the internal logic as well.

natemoo-re (commented):
Looking great so far! Left a few comments—happy to clarify if anything doesn't make sense.

.changeset/tangy-mirrors-hug.md
43081j: let's discuss on discord 👍 

i think that's fine though

davidwarrington:
I'm really liking this so far, but I have a few thoughts

- Should the input support min and max dates?
- When you interact with the input for the first time, the other segments default to a value of `00`. Since month and date can't be `00` I'm not sure if this makes sense. A benefit of the current implementation is it requires the user to select their own month and date still, rather than being able to accidentally submit `01` before they've had an opportunity to change it, but if that's the intention, I'd expect the segments to remain `__` until updated manually.
- The expected input format isn't obvious. Given `__/__/____` how do I know which segment represents month and which represents date?
- The input formats use `/` as a separator, but the rendered output uses `-`.  This can be ignored if the input switches to use `Date` though.

43081j:
it'd be good to render `dd / mm / yyyy` for example and, as you fill it in, it clears

so it goes like this:

> dd / mm / yyyy (all of these are grey/subtle)
> 10 / mm / yyyy (10 is white, mm/yyyy are grey still)

etc

dreyfus92:
Thanks everyone for your feedback so far 😄

dreyfus92:
i've tested this locally and i don't get any errors, which is weird.

I think the CI is failling due to the internal value i set with`new Date(iso)` where iso is "2025-01-15". in js that is UTC midnight. so the ui uses local `getFullYear()` / `getMonth()` /  `getDate()` to show the date.

so:
- CI (UTC): 2025-01-15T00:00:00Z → local day = 15.
- my machine: same instant → local day = 14.

also maybe snapshots were created on my local env (14), while CI (UTC) produced 15, so the snapshot comparison failed.

i might need to think about fixing this in the upcoming days.

packages/core/src/prompts/date.ts
```diff
@@ -0,0 +1,373 @@
...
+	inlineError = '';
+
+	#refreshFromSegmentValues() {
+		const display = this.#config.format(this.#segmentValues);
+		this._setUserInput(display);
+		this._setValue(segmentValuesToDate(this.#segmentValues) ?? undefined);
+	}
+
+	get cursor() {
+		// Convert segment cursor to absolute position for backward compatibility
```
natemoo-re: Is backward compatibility needed?

packages/core/src/prompts/date.ts
```diff
@@ -0,0 +1,373 @@
...
+		// Calculate start position by summing previous segments' lengths + separators
+		let start = 0;
+		for (let i = 0; i < this.#cursor.segmentIndex; i++) {
+			start += this.#config.segments[i].len + 1; // +1 for separator
+		}
+		return start + this.#cursor.positionInSegment;
+	}
+
+	/** Get current segment cursor position */
```
natemoo-re:
```suggestion
```

packages/core/src/prompts/date.ts
```diff
@@ -0,0 +1,373 @@
...
+	minDate?: Date;
+	maxDate?: Date;
+}
+
+export default class DatePrompt extends Prompt<Date> {
+	#config: DateFormatConfig;
+	#segmentValues: DateParts;
+	#minDate: Date | undefined;
+	#maxDate: Date | undefined;
+	/** Segment-based cursor: { segmentIndex, positionInSegment } */
```
natemoo-re:
```suggestion
```

packages/core/src/prompts/date.ts
```diff
@@ -0,0 +1,373 @@
...
+		if (!seg) return undefined;
+		// Ensure positionInSegment is within bounds
+		this.#cursor.positionInSegment = Math.max(
+			0,
+			Math.min(seg.len - 1, this.#cursor.positionInSegment)
+		);
+		return { segment: seg, index };
+	}
+
+	#onCursor(key?: string) {
```
natemoo-re: Feels like this could be broken down so we call out to methods like `#incrementCurrentSegment`, `#decrementCurrentSegment`, `#moveCursorNext`, `#moveCursorPrevious`

packages/core/src/prompts/date.ts
```diff
@@ -0,0 +1,373 @@
...
+	const val = (s: string) => Number.parseInt((s || '0').replace(/_/g, '0'), 10) || 0;
+	return {
+		year: val(parts.year),
+		month: val(parts.month),
+		day: val(parts.day),
+	};
+}
+
+function daysInMonth(year: number, month: number): number {
+	return new Date(year || 2000, month || 1, 0).getDate();
```
natemoo-re: 2000 was a leap year, so maybe not the best default?

43081j:
Also just finished reading through it. I think I have a few ideas for simplifying the logic so will open a follow up when I get time

PR #457, opened 2026-02-04
chore(prompts): destruct `limitOption` param for better code readability

Simek:
# Why

While going through the codebase I have spotted that `limitOption` param are partially destructed, and partially used directly by key reference. It seems to me that it makes code a bit harder to read, and understand default values at a glance.

# How

Destruct `limitOption` param fully in the function definition for better code readability, apply small tweaks to type definition (combining `[]` with `Array<>` for the same type looked really odd).

### DIFF
--- a/packages/prompts/src/limit-options.ts
+++ b/packages/prompts/src/limit-options.ts
@@ -13,7 +13,7 @@ export interface LimitOptionsParams<TOption> extends CommonOptions {
 }
 
 const trimLines = (
-	groups: string[][],
+	groups: Array<string[]>,
 	initialLineCount: number,
 	startIndex: number,
 	endIndex: number,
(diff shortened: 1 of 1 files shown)


packages/prompts/src/limit-options.ts
```diff
@@ -1,20 +1,19 @@
...
 	cursor: number;
 	style: (option: TOption, active: boolean) => string;
+	maxItems?: number;
 	columnPadding?: number;
 	rowPadding?: number;
 }
 
 const trimLines = (
-	groups: Array<string[]>,
+	groups: string[][],
```
43081j: we purposely prefer `Array<T>` as it is more readable and greppable than `T[][]`

```suggestion
	groups: Array<string[]>,
```

PR #460, opened 2026-02-05
fix: respect withGuide option in password and path prompts

### LINKED ISSUE #458: [Bug] incorrect spacing in `password` and `path` validation state with `withGuide: false`
**Environment**

- OS: macOS and Windows
- Node Version: 24.13.0
- Package: `@clack/prompts`
- Package Version: `main`

**Describe the bug**

User input and warning message in invalid state include whitespace characters (spacing) on the left side.

**To Reproduce**
1. Add `withGuide: false` to all prompts in `example/basic`
2. Run `pnpm dev`
3. Enter not valid input in `password` and `path` steps

**Expected behavior**

User input and warning message in invalid state does not include whitespace characters (spacing) on the left side.

**Additional Information**

Image

Image

kaigritun:
## Summary

When `withGuide: false` is set, the `password` and `path` prompts incorrectly included leading whitespace in:
- User input text
- Validation error messages

### `packages/prompts/src/password.ts`
- Makes indent spacing conditional on `hasGuide` in the error state
- When `hasGuide: false`, no leading whitespace is added to masked text or error message

### `packages/prompts/src/autocomplete.ts` (used by `path`)
- Imports `settings` from `@clack/core`
- Adds `hasGuide` check using `opts.withGuide ?? settings.withGuide`
- Makes all guide bar prefixes and spacing conditional on `hasGuide`
- Adjusts `columnPadding` for options based on guide presence

## Testing

All 579 tests pass (`pnpm build && pnpm test`).

Fixes #458


kaigritun:
Updated to include the indent in the prefix as suggested. All tests passing with updated snapshots.

kaigritun:
Fixed both issues:

1. **Simplified footers** - removed the `hasGuide` conditional since `guidePrefix` is already empty when there's no guide
2. **Fixed double-dim** - the instructions array already had `color.dim()` applied, so wrapping the join in another `color.dim()` was causing the `[2m[2m` issue in snapshots. Now only dim the separator: `instructions.join(color.dim(' • '))`

All tests passing with updated snapshots.

kaigritun:
Fixed the trailing spaces issue - moved the indent back into the label/userInputText rather than the prefix, so there are no trailing spaces when the value is empty (e.g., cancel without typing anything). Snapshots updated.

kaigritun:
Understood - I can revert to `color.dim(instructions.join(' • '))` to minimize snapshot changes. However, the instructions array items already have `color.dim()` applied to certain parts (e.g., `color.dim('↑/↓')`), which is why I changed it to only dim the separator. This avoids double-dim escape codes in the output.

Would you prefer:
1. Keep current approach (`instructions.join(color.dim(' • '))`) - cleaner escape codes but requires snapshot updates
2. Revert to `color.dim(instructions.join(' • '))` - no snapshot changes but has nested dim codes

Happy to go either way!

43081j:
sorry lost track of the comments in here and the comments in the code, so had missed your previous explanation

the instructions are already partially dimmed and then the outer set is dimmed as a whole. this does seem like nonsense, since what was the point in dimming the individual parts if the only consumer of this `instructions` is going to dim it all anyway?

i think we should remove the outer `dim` entirely

kaigritun:
Done - removed the outer `dim()` entirely. The individual instruction parts retain their styling (`color.dim('↑/↓')`, etc.) while the separator and descriptive text are now normal brightness.

biome.json
```diff
@@ -1,7 +1,7 @@
 {
   "$schema": "https://biomejs.dev/schemas/2.1.2/schema.json",
   "vcs": { "enabled": false, "clientKind": "git", "useIgnoreFile": false },
-  "files": { "ignoreUnknown": false, "includes": ["**", "!**/dist/**"] },
+  "files": { "ignoreUnknown": false, "includes": ["**", "!**/dist"] },
```
43081j: last comment i think - why wasn't this working before?

kaigritun:
Good catch, removed the lockfile. 👍

43081j:
if you catch up from `main` the CI will pass (just discard your changes to `utils.test.ts`)

there's one comment left on the biome config change too

kaigritun:
Rebased from main and reverted the biome.json change - you're right, it wasn't needed. Should be good now.

kaigritun:
Removed the @ts-expect-error comments per feedback.

@43081j @xandnand - there's conflicting feedback on the outer dim: James suggested removing it entirely, but xandnand thinks we should add it back. Could you align on the preferred approach? Happy to implement whichever direction you decide.

43081j (approved):

43081j:
did you mean to at-mention someone else? 👀  not sure who xandnand is

anyhow, i'm happy with the change in look. we can reach out to some large users like astro before the next release to check all is good if necessary too. 

not sure why we were dimming the entire line before - more than likely it was a change by me a while ago and i hadn't realised

YoungElPaso:
Is this the AI PR if so wanna flag it somehow?

43081j:
This is indeed an ai PR. 

It seemed obvious from the first post but the change is something we'd been meaning to do anyway.

Given it is just rep farming maybe it is worth flagging the account, but not necessarily the PR since it is a change I was already planning on making that same day.

PR #463, opened 2026-02-09
Add wraparound to autocomplete

### LINKED ISSUE #462: [Bug] No wraparound for autocomplete
Environment

OS: macOS
Node Version: v22
Package: @clack/core
Package Version: v1.0.0

Describe the bug

The AutocompletePrompt does not wrap the cursor when pressing up at the top or down at the bottom of the options list. The cursor clamps to the bounds instead of wrapping around.

In AutocompletePrompt, the cursor is clamped:

https://github.com/bombshell-dev/clack/blob/ba1072166a3bdb11d6389a3a1c881cfc3ce985d1/packages/core/src/prompts/autocomplete.ts#L145-L149

Whereas, SelectPrompt and GroupMultiSelectPrompt wrap around using moveCursor:

https://github.com/bombshell-dev/clack/blob/ba1072166a3bdb11d6389a3a1c881cfc3ce985d1/packages/core/src/prompts/select.ts#L33-L43

https://github.com/bombshell-dev/clack/blob/ba1072166a3bdb11d6389a3a1c881cfc3ce985d1/packages/core/src/utils/cursor.ts#L6-L9

note: `clampedCursor` as a variable name is a bit misleading here as this is where the wraparound logic resides.

To Reproduce

Create an autocomplete prompt with multiple options
Press the down arrow until you reach the last option
Press down again — the cursor stays on the last item instead of wrapping to the first
Similarly, pressing up on the first item does not wrap to the last

Expected behavior

The cursor should wrap around, matching the behavior of SelectPrompt and GroupMultiSelectPrompt. Pressing down on the last item should move focus to the first item, and pressing up on the first item should move focus to the last item.

Additional Information

Happy to open a PR for this if the maintainers agree on the expected behaviour.

djgrant:
Adds wraparound behaviour to autocomplete and multiselect autocomplete. 

Tested both examples (`pnpm --filter @example/basic exec jiti ./autocomplete.ts`) and the wraparound behaviour works as expected tranistioning from first to last on up arrow, and last to first on down arrow.

Resolves #462.

### DIFF
--- a/packages/core/src/prompts/autocomplete.ts
+++ b/packages/core/src/prompts/autocomplete.ts
@@ -6,6 +6,7 @@ import { findCursor } from '../utils/cursor.js';
 interface OptionLike {
 	value: unknown;
 	label?: string;
+	disabled?: boolean;
 }
 
 type FilterFunction<T extends OptionLike> = (search: string, opt: T) => boolean;


43081j:
CI is failing because `T` doesn't extend `{disabled?: boolean}`

you should be able to just update autocomplete's `OptionLike` to have this optional property

Issue #467, opened 2026-02-11
[Request] Sorted autocomplete results.


HamishWHC:
**Is your feature request related to a problem? Please describe.**
I am re-implementing a script to search for an item and then perform some actions on it. The original script used fzf (a fuzzy search bash utility). This had the advantage of the _closest_ match would appear at the top of the list. I would like to achieve the same with clack + fuse.js or a similar fuzzy search library, but clack only supports filtering, rather than applying an ordering to the results.

**Describe the solution you'd like**
I'd like to see an alternative filter option that allows for filtering and changing the order of the items returned (notably, if this filter option allows a promise to be returned, this could also be used to implement async autocompletion).

**Describe alternatives you've considered**
For now I'm just leaving the matches unordered. I could try using another library, but clack is the nicest [ie]nquirer alternative I've found.

**Additional context**

HamishWHC:
Upon further investigation seems this is likely doable by using a function for `options`? This is great, but ideally should be documented somewhere.

43081j:
indeed, autocomplete options can be a getter function where you can then sort them.

we should add this to the docs 👍 

ill rename this issue

PR #471, opened 2026-02-22
feat(core, prompts): add async autocomplete options

mateussilva92:
This PR enhances the autocomplete prompt to support async option functions.

## Fixes:
#467 

### Autocomplete:
- Autocomplete options can now be provided as an async function;
- Request cancellation via AbortController;
- Stale responses are safely ignored to prevent race conditions;
- Proper loading state handling;

## Common/Helper:
- `unicodeOr` is now generic
- `isAsync` function was added.

## Backwards Compatibility:
- Existing sync configurations continue to work unchanged.
- No existing unit tests changed.

## Notes:
This PR has the feature itself fully developed. I didn't create any unit tests yet since I want to get feedback on my implementation. When I get an thumbs up on the implementation I'll add the unit tests / examples and remaining files.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -17,6 +17,13 @@ export { default as SelectKeyPrompt } from './prompts/select-key.js';
 export type { TextOptions } from './prompts/text.js';
 export { default as TextPrompt } from './prompts/text.js';
 export type { ClackState as State } from './types.js';
-export { block, getColumns, getRows, isCancel, wrapTextWithPrefix } from './utils/index.js';
+export {
+	block,
+	getColumns,
+	getRows,
+	isAsync,
+	isCancel,
+	wrapTextWithPrefix,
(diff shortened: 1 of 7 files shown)


43081j:
this is a very large change for something we haven't seen much demand for. also note that this doesn't fix #467, it just adds async on top of what that is asking to document.

i very much appreciate the work that has gone into this, but am tempted to leave it sync-only until we see some real demand for it. otherwise it is a slippery slope to introducing a lot of extra code for edge case features.

@dreyfus92 thoughts?

dreyfus92:
i agree. i'd rather keep it as it is until we see a real demand, this is a lot of surface for an edge case. a couple of things for context: #467 is asing for a reordering filter, not async, and that's already doable with the options getter reading `this.userInput`. so the fix there is docs, happy to review a pr for that.

PR #478, opened 2026-03-04
fix(prompts): correct spinner clearing by tracking previous rendered output

escral:
## PR Description

First of all, sorry for not opening an issue before submitting this PR.

## The issue

When the spinner message is shorter than the terminal width, but becomes longer after the `"..."` suffix added by the `"dots"` indicator, the previous frame is not cleared correctly.

Currently the clearing logic calculates the number of lines based on `_prevMessage`. However, `_prevMessage` is **not the actual rendered output**. The rendered output also includes:

- the spinner frame
- indicator dots (`.` / `..` / `...`)
- timer text (when using `indicator: "timer"`)
- ANSI styling

Because `_prevMessage` does not include these elements, the number of wrapped lines can be calculated incorrectly. As a result, the spinner sometimes fails to fully clear the previous frame.

## How to reproduce

A reproduction script was added:

`examples/basic/spinner-wrap.ts`

Manual reproduction steps:

1. Start a spinner with a message whose length is close to the terminal width (e.g. terminal columns minus ~3 characters for the spinner frame).
2. Once the `"dots"` indicator appends `.` → `..` → `...`, the message becomes longer than the terminal width.
3. The previous spinner frame is not fully cleared, leaving part of the previous line visible.

## Fix

Instead of tracking `_prevMessage`, the fix tracks the **previous rendered output**.  
This ensures the clearing logic uses the exact string that was written to the terminal, producing correct line calculations when wrapping occurs.

### Current behavior

https://github.com/user-attachments/assets/9f2e6f50-06c4-4b9c-9ccc-316d8ca1f638

### With fix

https://github.com/user-attachments/assets/a52ecb8c-0fdb-49c6-98a7-9ac7c204d2e9

### DIFF
--- a/.changeset/hungry-drinks-raise.md
+++ b/.changeset/hungry-drinks-raise.md
@@ -1,0 +1,5 @@
+---
+"@clack/prompts": patch
+---
+
+Fix spinner clearing by tracking the previously rendered output instead of message.


43081j:
we're actually in the middle of reworking the spinner to be a `Prompt` like all the other prompts. this becomes a non-issue then i think because we'll be using the same underlying diff rendering as other prompts.

so im not sure if we should merge this. maybe instead, we should try get the spinner rework over the line

will wait and see what @dreyfus92 thinks too

dreyfus92 (commented):
Thank you @escral for taking the time to work on this, we're planning to rework the spinner so it uses the same Prompt base class and diff rendering as our other prompts (text, select, etc.) in that design, like @43081j  stated, we always diff the full rendered frame, so this clearing bug doesn't arise and the current clearing path goes away. Because the rework would replace this code path, we're hesitant to merge this fix as-is. We'd rather not add code we know we'll remove soon, we'd like to prioritise getting the spinner rework over the line. 

We really appreciate the fix and the repro, we'll keep it in mind when we do the rework 💜

escral:
Sure, thanks for the quick response! Looking forward to the reworked spinner.

43081j:
ill close this and open a draft with the spinner rework 👍 

will try find time in the next week to finish it off

PR #479, opened 2026-03-05
feat: rework spinner to use `Prompt` api

43081j:
This reworks the spinner to be based on the `Prompt` class and have a `core` prompt like the others do.

It should solve a few rendering bugs/inconsistencies.

DRAFT until i have chance to fix tests up


packages/core/src/prompts/spinner.ts
```diff
@@ -0,0 +1,183 @@
...
+
+		this.#reset();
+		this.#silentExit = silent === true;
+		this.#exitCode = exitCode ?? 0;
+
+		if (msg !== undefined) {
+			this.#message = msg;
+		}
+
+		this.state = 'cancel';
```
natemoo-re, line 77: Should we do `this.state = 'submit'` when the `exitCode === 0`? Currently this would treat every exit as `cancel`.

natemoo-re (commented):
Code looks good! One thing that regresses here is when pressing `Enter` we're getting frames spammed to the console. Seems like we could add special handling for this—I remember fixing that with the previous implementation

Image

.changeset/empty-bananas-melt.md
```diff
@@ -0,0 +1,6 @@
+---
+"@clack/prompts": patch
+"@clack/core": patch
```
natemoo-re: Should be okay as an internal refactor, but since this technically alters the raw-mode behavior we might want to do this in a minor to be safe?

dreyfus92 (approved):

43081j:
@dreyfus92 @natemoo-re FYI this is blocked because it has brought the ol' windows bug back.

we still haven't figured out how to fix that bug properly in main. the proposed fix in main introduces the _other_ windows bug.

so this branch has pretty much the same issues until i can figure out the right solution.

PR #481, opened 2026-03-06
chore: externalize fast-string-width & fast-wrap-ansi

hyperz111:
Because we now on ESM-only package, maybe we can externalize these dependencies:
- `fast-string-width`
- `fast-wrap-ansi`

### DIFF
--- a/.changeset/afraid-donkeys-sin.md
+++ b/.changeset/afraid-donkeys-sin.md
@@ -1,0 +1,6 @@
+---
+"@clack/prompts": minor
+"@clack/core": minor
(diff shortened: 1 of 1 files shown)


43081j:
Can you add a change set? 

Generally looks good to me, and means we can lean into npm de-dupe since these two packages are quickly growing in usage

hyperz111:
> Can you add a change set? 

I will do it tomorrow (I'm on GMT+7)

hyperz111:
I think this is can be a minor change.

Issue #477, opened 2026-03-04
[Bug] path directory cant direct select default value


lgl017temp:
When I have a default value for the directory, pressing Enter still selects the first item in the list instead of the current value.I must delete a character to select the default directory.

PR #484, opened 2026-03-09
fix(prompts): submit initial directory value in path prompt

dreyfus92:
this pr fixes path directory mode so when an existing directory is provided as initialValue, pressing enter immediately submits that current directory instead of the first child option.

closes: #477


43081j:
@dreyfus92 what do you think of these changes i added?

basically `/foo` will show `/foo` (if it is a dir), but `/foo/` will show child directories of `/foo`.

without this change, there's no longer a way to traverse into `/foo` other than knowing the name of a child by memory

dreyfus92:
this is clever @43081j, and nothing breaks so ig it works! 😄

PR #489, opened 2026-03-16
fix(core): skip double-filtering when options is a function

### LINKED ISSUE #488: [Bug] `filter` + `options()` double-filtering issue for autocomplete
### Environment

- OS: macOS
- Node: 25.8.1
- Package: `@clack/prompts`
- Package Version: 1.1.0

### Describe the bug

When `options` is a function, clack internally calls it to get the candidate
list, then still runs its own `filter` function over that result. This means any
pre-filtered / pre-sorted list returned by `options()` gets filtered a second
time by clack's default substring matcher — silently dropping items that a
custom algorithm (e.g. fuzzysort) considered valid matches.

So the user is forced to include `filter: () => true` to workaround this ergonomic pitfall.

### To Reproduce

https://stackblitz.com/edit/node-ynqghq1u?file=index.js

run the stackblitz script with: `node index.js`

Notice: ❌ nothing is being listed

```txt
❯ node index.js
│
◆  Select a package
│
│  Search: vte█ (0 matches)
│  No matches found
│  ↑/↓ to select • Enter: confirm • Type: to search
```

repeat with `node index.js --filter`

Notice: ✅ correct list with `vite` at the top

```txt
❯ node index.js --filter
│
◆  Select a package
│
│  Search: vte█
│  ○ vite
│  ○ vitest
│  ● vitepress
│  ↑/↓ to select • Enter: confirm • Type: to search
```

### Expected behavior

When `options` is a function (i.e. the caller manages filtering), clack should
not apply an additional `filter` pass on the returned list.

### Additional Information

You may say _'why dont u use fuzzysort in the filter?'_

If I did that, I’d lose the ranking: `vite` would move from first to last.

```js
filter(search, option) {
  if (!search) {
    return true
  }
return fuzzysort.go(search, [option.value]).length > 0
},
options: packages.map(name => ({ label: name, value: name }))
```

```txt
❯ node index.js
│
◆  Select a package
│
│  Search: vte█ (3 matches)
│  ● vitepress
│  ○ vitest
│  ○ vite
```

### Related Prs:

- https://github.com/bombshell-dev/clack/pull/288
  - Introduced autocomplete. filter always runs over options() result
- https://github.com/bombshell-dev/clack/pull/444
  - Exposed filter as a option

---

PS: The point of this issue report is that I found it counterintuitive to
include `filter: () => true` as a workaround for this issue. I hope you agree or
can provide a documentation example so that more people won’t encounter this
pitfall.

brick-pixel:
## What

When `options` is provided as a function, the autocomplete prompt no longer applies the default filter on top of the function's already-filtered results.

## Why

As reported in #488, when `options` is a function (e.g. using fuzzysort for custom ranking), clack's default substring filter runs on the returned list, silently dropping valid matches. This forces users to add `filter: () => true` as a workaround.

**Root cause:** In `#onUserInputChanged`, `this.#filterFn` is always applied to the result of `this.options`, regardless of whether `options` is a static array or a dynamic function.

## Changes

**File:** `packages/core/src/prompts/autocomplete.ts`

- Added `#hasCustomFilter` field to track whether a custom `filter` was provided
- When `options` is a function and no custom `filter` was provided, the returned options are used as-is
- When both `options` (function) and `filter` are provided, the filter is still applied as before

## Behavior Matrix

| `options` | `filter` | Behavior |
|-----------|----------|----------|
| Array | Not provided | Default filter applied ✅ (unchanged) |
| Array | Custom function | Custom filter applied ✅ (unchanged) |
| Function | Not provided | **No filter applied** ✅ (fixed) |
| Function | Custom function | Custom filter applied ✅ (unchanged) |

## Verification

- All existing tests pass (630 tests, 0 failures)
- Lint and format checks pass (biome)

Fixes #488

### DIFF
--- a/packages/core/test/prompts/autocomplete.test.ts
+++ b/packages/core/test/prompts/autocomplete.test.ts
@@ -197,4 +197,59 @@ describe('AutocompletePrompt', () => {
 		expect(instance.selectedValues).to.deep.equal([]);
 		expect(result).to.deep.equal([]);
 	});
+
+	test('options function skips default filter when no custom filter provided', () => {
+		const optionsFn = function (this: any) {
+			const input = this.userInput || '';
+			if (!input) return testOptions;
+			return testOptions.filter((opt) => opt.value.startsWith(input));
+		};
+
+		const instance = new AutocompletePrompt({
+			input,
+			output,
+			render: () => 'foo',
+			options: optionsFn,
+		});
+
+		instance.prompt();
+
+		// Type 'b' - options function returns only 'banana'
+		input.emit('keypress', 'b', { name: 'b' });
+
+		// Should use the options function result as-is, no additional filtering
+		expect(instance.filteredOptions).toEqual([
+			{ value: 'banana', label: 'Banana' },
+		]);
+	});
+
+	test('options function applies custom filter when both are provided', () => {
+		const optionsFn = function (this: any) {
+			return testOptions;
+		};
+
+		const customFilter = (search: string, opt: { value: string; label?: string }) => {
+			return opt.value.includes(search);
+		};
+
+		const instance = new AutocompletePrompt({
+			input,
+			output,
+			render: () => 'foo',
+			options: optionsFn,
+			filter: customFilter,
+		});
+
+		instance.prompt();
+
+		// Type 'an' - custom filter should run on top of options function result
+		input.emit('keypress', 'a', { name: 'a' });
+		input.emit('keypress', 'n', { name: 'n' });
+
+		// Custom filter: 'an' matches banana and orange
+		expect(instance.filteredOptions).toEqual([
+			{ value: 'banana', label: 'Banana' },
+			{ value: 'orange', label: 'Orange' },
+		]);
+	});
 });


43081j:
for some context (cc @dreyfus92 too), we (or at least i) were aware of this. **it is not a bug**, it was an active decision that if you want to do your own filtering, you need to pass a noop filter until we get around to deciding how to solve it in a better way.

so we're just now at the point of "deciding how to solve it in a better way".

just wanted to make that clear so we're aware this is not a bug we're fixing. it is a behavioural change and a `feat`.

not all custom options functions do their own filtering, which is why we didn't simply turn it off like in this PR.

i feel like it is a rare case, though, to have the need for an options function and _not_ do your own filtering. so i think i am happy to change this. just make sure you update the PR title etc to `feat`

43081j (approved):

43081j (changes requested):
ah lastly can you add tests for the new branches?

- setting an options function doesn't apply the default filter
- setting a filter and an options function works (i.e. both run)

brick-pixel:
Thanks for the context! Updated the PR:

- Title changed to `feat`
- Changeset bumped from `patch` to `minor` since this is a behavioral change

Makes sense that not all custom options functions do their own filtering. With this change, `filter` still works normally when explicitly provided alongside a function `options` — only the implicit default filter is skipped.

brick-pixel:
Added the tests — one for options function without filter (default filter skipped) and one with both options function and custom filter (filter still runs). All 98 core tests passing.

PR #490, opened 2026-03-17
improve path component usage

lgl017temp:
improve path component
1.tab can input selected value
2.put current dict into options
3.allow not exists

To minimize changes, hidden internal API `_clearUserInput` and `_setUserInput` was used; a refactoring is recommended.

### DIFF
--- a/packages/prompts/package.json
+++ b/packages/prompts/package.json
@@ -53,7 +53,7 @@
     "test": "vitest run"
   },
   "dependencies": {
-    "@clack/core": "workspace:*",
+    "@clack/core": "file:../core",
     "fast-string-width": "^1.1.0",
     "fast-wrap-ansi": "^0.1.3",
     "sisteransi": "^1.0.5"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -58,6 +58,18 @@ importers:
         specifier: ^1.0.0
         version: 1.1.1
 
+  examples/testpath:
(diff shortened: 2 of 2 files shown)


43081j:
i think these need more discussion before a PR.

as far as I understand, this is what you've added:

- `exists` is a way to allow the user to enter paths which don't exist
- you can now press tab to append the selected path to the user input, and continue typing
- add the current user input to the list of results (? not sure why)

but already these are 3 distinct changes which each need to be discussed before we decide if to do them or not.

lgl017temp:
> i think these need more discussion before a PR.
> 
> as far as I understand, this is what you've added:
> 
> * `exists` is a way to allow the user to enter paths which don't exist
> * you can now press tab to append the selected path to the user input, and continue typing
> * add the current user input to the list of results (? not sure why)
> 
> but already these are 3 distinct changes which each need to be discussed before we decide if to do them or not.



1. Sometime I need select a output path, it may be not exists, I want accept a don't exist dir then call mkdir, better than I create it in explore manually.
2. When I want into a dir, I must input full name manually. So I want provide a quickly way, like terminal.
3. #477 At first I just want fix it, then you fix it too. But when I test it, I found you not use sep, it cause not list sub dir when ends with "\\". And When combined with the second point, i need add "\\" manually. So I keep my solution.

lgl017temp:
Close to prevent subsequent push from being incorrectly PR

PR #496, opened 2026-03-23
feat(autocomplete): do not set default filter for options getters

### LINKED ISSUE #488: [Bug] `filter` + `options()` double-filtering issue for autocomplete

43081j:
Basically:

- If `options` is a function, do not set the default filter
- If a filter is explicitly set, keep it
- If `options` is not a function, set the default filter

Closes #488.
Closes #489.

### DIFF
--- a/.changeset/silent-sides-call.md
+++ b/.changeset/silent-sides-call.md
@@ -1,0 +1,5 @@
+---
+"@clack/core": patch
+---
+
+Only apply autocomplete default filter if it has been explicitly set or if options is not a getter.
--- a/packages/core/src/prompts/autocomplete.ts
+++ b/packages/core/src/prompts/autocomplete.ts
@@ -71,7 +71,7 @@ export default class AutocompletePrompt<T extends OptionLike> extends Prompt<
 	focusedValue: T['value'] | undefined;
 	#cursor = 0;
 	#lastUserInput = '';
-	#filterFn: FilterFunction<T>;
+	#filterFn: FilterFunction<T> | undefined;
 	#options: T[] | (() => T[]);
 	#placeholder: string | undefined;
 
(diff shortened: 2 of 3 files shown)


43081j:
@dreyfus92 the other PR went stale and got flagged as bot-like, which may explain the staleness

either way, here it is from scratch so we can get it landed

PR #504, opened 2026-04-02
chore: remove unused debug key

hyperz111:
`debug` is unused (maybe), so i remove it.


43081j:
can you add a changeset?

this is technically breaking in that someone could have it in their source, but it is a bug that it is there at all. so i think im happy it being in a patch

hyperz111:
this should be a minor change instead, i think.

43081j (approved):

PR #509, opened 2026-04-08
fix(@clack/prompts): handle carriage return output in taskLog

colinaaa:
## Summary
- handle `taskLog` raw message updates that use carriage returns so spinner frames do not accumulate
- add regression coverage for carriage-return spinner output and `error()` rendering behavior
- add a changeset for the `@clack/prompts` patch release

## Test Plan
- `cd packages/prompts && pnpm test`

### DIFF
--- a/.changeset/lovely-radios-joke.md
+++ b/.changeset/lovely-radios-joke.md
@@ -1,0 +1,5 @@
+---
+"@clack/prompts": patch
+---
+
+Fix `taskLog` raw message handling so carriage-return spinner updates do not accumulate repeated frames.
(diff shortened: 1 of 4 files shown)


43081j:
while this may be a nice fix to have, its not the way we should fix this.

the spinner has similar issues in that it is a "raw" prompt (doesn't inherit `Prompt` and implements its own rendering logic).

going forward, the spinner is going to be refactored to use the same rendering logic as every other prompt. The task log prompt should do the same, and that will solve this problem at the same time.

an issue goes a long way here as not all this direction is written down in one place, so you couldn't really have known that refactor was in progress.

colinaaa:
@43081j thanks for your review! I'll create an issue instead.

PR #512, opened 2026-04-15
chore(deps): update deps in core and prompts packages

dreyfus92:


43081j (approved):
it builds! fair enough 😂 

it is probably worth manually trying it out a bit too. do some weird stuff like a box with a lot of colour codes in it, or a box with a link in it is a good one

gameroman:
> it is probably worth manually trying it out a bit too. do some weird stuff like a box with a lot of colour codes in it, or a box with a link in it is a good one

Should I try adding some tests with snapshots?

gameroman:
I've just tested and the following fails on the main branch, I will test on this branch now

```js
const longUrl = 'https://example.com/very/long/path/that/exceeds/typical/terminal/width';
const link = `\x1b]8;;${longUrl}\x1b\\click\x1b]8;;\x1b\\`;
prompts.box(`Link: ${link}`, 'Long URL', { input, output, width: 'auto' });
```

 Tested, it passes on this branch, see https://github.com/gameroman/clack/commit/71efa6760ef8598a8d2c9083a5858625e51fefe3

43081j:
we don't need new tests for it, i just want us to manually try out a few different cases to be sure its all still fine.

sounds like it is 👍

gameroman (approved):

PR #513, opened 2026-04-20
feat(core): support number key jump in select prompt

### LINKED ISSUE #475: select: support number key to jump to option
## Feature Request

Allow pressing number keys (1-9) to jump to and select the corresponding option in the `select` prompt.

### Current Behavior

Users must use arrow keys to navigate options and press Enter to confirm.

### Proposed Behavior

When a `select` prompt has numbered options, pressing a number key (e.g. `1`, `2`, `3`) should immediately jump to and select that option 

This is especially useful for short lists (< 10 items) where pressing a single key is much faster than arrow-key navigation.

### Example

```typescript
const answer = await select({
  message: 'Pick a market',
  options: [
    { label: 'Market 1' },
    { label: 'Market 2' },
    { label: 'Market 3' },
  ],
})
// Pressing "2" immediately selects "Market 2"
```

rodindev:
## What does this PR do?

Press keys 1–9 in the select prompt to jump the cursor to that option. Out-of-range digits are ignored; if the target is disabled, the cursor skips forward to the next enabled option (same as the down-arrow). Enter still submits.

Closes #475 

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

### DIFF
--- a/.changeset/clever-tigers-jump.md
+++ b/.changeset/clever-tigers-jump.md
@@ -1,0 +1,5 @@
+---
+"@clack/core": minor
+---
+
+Press number keys 1–9 in the select prompt to jump to the corresponding option.
--- a/packages/core/src/prompts/select.ts
+++ b/packages/core/src/prompts/select.ts
@@ -43,5 +43,15 @@ export default class SelectPrompt<T extends { value: any; disabled?: boolean }> 
 			}
 			this.changeValue();
 		});
+
+		this.on('key', (char, key) => {
+			if (key.ctrl || key.meta) return;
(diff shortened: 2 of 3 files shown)


43081j:
We (maintainers) need to discuss the linked issue before we land something like this. Happy to leave this open until we decide though.

if we accept every suggestion, its a quick way to end up bloating clack. so we do need to choose carefully which features should be accepted.

ill report back to this PR once we've discussed

PR #515, opened 2026-04-21
deps: update pnpm and align node types version

ghostdevv:
## What does this PR do?

Updates pnpm to v10 as it works on Node 20, as well as downgrading the `@types/node` to use v20 to align

## Type of change

- [ ] Bug fix
- [ ] Feature
- [x] Refactor (no behavior change)
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [ ] I have added a changeset

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,13 +17,13 @@
   "devDependencies": {
     "@biomejs/biome": "^2.1.2",
     "@changesets/cli": "^2.29.5",
-    "@types/node": "^24.1.0",
+    "@types/node": "^20.19.39",
     "jsr": "^0.13.4",
     "knip": "^5.62.0",
     "typescript": "^5.8.3",
     "unbuild": "^3.6.0"
   },
-  "packageManager": "pnpm@9.14.2",
+  "packageManager": "pnpm@10.33.0",
   "volta": {
     "node": "20.18.1"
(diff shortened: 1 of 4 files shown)


43081j:
can you explain the downgrade of `@types/node`?

are you trying to match the types to the min node version? im not sure that makes the most sense since its a _minimum_ (i.e. locally its unlikely we have that). plus it means we now have to remember not to update it whenever we update our dev dependencies.

i think the latest is fine and we'd know if not, since the build would fail, and we want to be able to run clack in all modern nodes

ghostdevv:
> are you trying to match the types to the min node version? im not sure that makes the most sense since its a minimum (i.e. locally its unlikely we have that). plus it means we now have to remember not to update it whenever we update our dev dependencies.

Yea the idea is that it's pinned so you know through types when you're using something that's too new. Ideally you'd be running the same Node version as your minimum support locally, then tests can do the higher ups. However, volta/nvm is only currently configured in the root so that may not happen so then you'd know because the types'd fail. I know it's a bit of a pain when you update deps but it's been worth it in my experience, however let me know if you'd like the change removed :pray:

43081j (approved):

PR #518, opened 2026-04-26
fix: split type-only and runtime exports

### LINKED ISSUE #470: isCancel runtime export regression in `@clack/prompts`
Short summary
- A mixed `type` + value re-export in `packages/prompts/src/index.ts` can cause some TypeScript emitters/bundlers to drop runtime exports (e.g. `isCancel`). This breaks consumers following the docs/examples.

Reproduction (minimal)
```ts
import { isCancel, text } from '@clack/prompts';
// depending on toolchain, `isCancel` may be undefined or throw at runtime
```

Impact
- Blocks "Getting Started" examples and any consumer relying on `isCancel` from `@clack/prompts`.
- High user friction; likely to break downstream CLIs and automation.

Fix (implemented locally)
- Split the mixed export into a `type`-only export and a separate value export:
  - export type { ClackSettings } from '@clack/core';
  - export { isCancel, settings, updateSettings } from '@clack/core';

PoC (observed result)
- `typeof isCancel => function`
- `typeof settings => object`
- `typeof updateSettings => function`

Patch (exact change applied)
```diff
- export { type ClackSettings, isCancel, settings, updateSettings } from '@clack/core';
+ export type { ClackSettings } from '@clack/core';
+ export { isCancel, settings, updateSettings } from '@clack/core';
```

jeffrey-takuma:
## Summary

Splits the mixed type/value re-export from `@clack/core` into separate type-only and runtime exports.

This avoids runtime exports such as `isCancel`, `settings`, and `updateSettings` being dropped by some TypeScript emitters or bundlers.

Closes #470

## Testing

- `pnpm build`
- `pnpm types`

No tests were added because this only separates type-only and runtime re-exports.

### DIFF
--- a/.changeset/honest-crews-go.md
+++ b/.changeset/honest-crews-go.md
@@ -1,0 +1,5 @@
+---
+"@clack/prompts": patch
+---
+
+Fix mixed type-only and runtime exports from @clack/core.
--- a/packages/prompts/src/index.ts
+++ b/packages/prompts/src/index.ts
@@ -1,4 +1,5 @@
-export { type ClackSettings, isCancel, settings, updateSettings } from '@clack/core';
+export type { ClackSettings } from '@clack/core';
+export { isCancel, settings, updateSettings } from '@clack/core';
 
 export * from './autocomplete.js';
 export * from './box.js';


43081j (approved):

gameroman:
Would it make sense to add a biome lint rule for this?

43081j:
i'd rather us not have to maintain custom rules. but if this ends up existing inside biome, i'd be up for turning it on

gameroman:
Pretty sure there is already a builtin rule for that

PR #534, opened 2026-05-09
fix: spaces and uppercase characters in multiline input

### LINKED ISSUE #529: [Bug] Can't type spaces or uppercase letters into a multiline prompt
**Environment**

- OS: macOS
- Node Version: v22.17.0
- Package: @clack/core
- Package Version: v1.3.0

**Describe the bug**
Can't type spaces or uppercase letters into a multiline prompt

**To Reproduce**
https://stackblitz.com/edit/node-853mspgi?file=index.js

- `node index.js`
- Try typing in capital letters and spaces

**Expected behavior**
I should be able to type in capital letters and spaces

**Additional Information**
There is also some odd behavior when backspacing at the start of the line

MattStypa:
## What does this PR do?

Fixes multiline prompt typing so spaces and uppercase characters are inserted correctly.

See #529 for details.

Changes
- Use a local arrow-key action set in multi-line prompt handling.
- Fix character insertion to preserve spaces and shifted uppercase input.
- Add regression tests for space insertion and shift-uppercase typing.

Closes #529 

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

### DIFF
--- a/packages/core/src/prompts/multi-line.ts
+++ b/packages/core/src/prompts/multi-line.ts
@@ -1,8 +1,11 @@
 import type { Key } from 'node:readline';
 import { styleText } from 'node:util';
 import { findTextCursor } from '../utils/cursor.js';
-import { type Action, settings } from '../utils/index.js';
 import Prompt, { type PromptOptions } from './prompt.js';
+
+const actions = ['up', 'down', 'left', 'right'];
+type Action = (typeof actions)[number];
+const actionSet = new Set(actions);
 
 export interface MultiLineOptions extends PromptOptions<string, MultiLinePrompt> {
 	placeholder?: string;
@@ -89,7 +92,7 @@ export default class MultiLinePrompt extends Prompt<string> {
 		this.#showSubmit = opts.showSubmit ?? false;
 
 		this.on('key', (char, key) => {
-			if (key?.name && settings.actions.has(key.name as Action)) {
+			if (key?.name && actionSet.has(key.name as Action)) {
 				this.#handleCursor(key.name as Action);
 				return;
 			}
@@ -118,7 +121,8 @@ export default class MultiLinePrompt extends Prompt<string> {
 				if (this.#showSubmit && this.focused === 'submit') {
 					this.focused = 'editor';
 				}
-				this.#insertAtCursor(char ?? '');
+				const casedChar = key.shift ? char.toUpperCase() : char;
+				this.#insertAtCursor(casedChar ?? '');
 				this._cursor++;
 			}
 		});
--- a/packages/core/src/prompts/multi-line.ts
+++ b/packages/core/src/prompts/multi-line.ts
@@ -1,11 +1,8 @@
 import type { Key } from 'node:readline';
 import { styleText } from 'node:util';
 import { findTextCursor } from '../utils/cursor.js';
+import { type ArrowKeyAction, settings } from '../utils/index.js';
 import Prompt, { type PromptOptions } from './prompt.js';
-
-const actions = ['up', 'down', 'left', 'right'];
-type Action = (typeof actions)[number];
-const actionSet = new Set(actions);
 
 export interface MultiLineOptions extends PromptOptions<string, MultiLinePrompt> {
 	placeholder?: string;
@@ -44,7 +41,7 @@ export default class MultiLinePrompt extends Prompt<string> {
 			this.userInput.slice(0, this.cursor) + char + this.userInput.slice(this.cursor)
 		);
 	}
-	#handleCursor(key?: Action) {
+	#handleCursor(key?: ArrowKeyAction) {
 		const text = this.value ?? '';
 		switch (key) {
 			case 'up':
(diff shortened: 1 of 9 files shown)


packages/core/src/prompts/multi-line.ts
```diff
@@ -1,9 +1,12 @@
 import type { Key } from 'node:readline';
 import { styleText } from 'node:util';
 import { findTextCursor } from '../utils/cursor.js';
-import { type Action, settings } from '../utils/index.js';
 import Prompt, { type PromptOptions } from './prompt.js';
 
+const actions = ['up', 'down', 'left', 'right'];
+type Action = (typeof actions)[number];
+const actionSet = new Set(actions);
+
```
gameroman: This can probably be added to `utils/settings` as `arrowKeys`

packages/core/src/prompts/multi-line.ts
```diff
@@ -118,7 +121,8 @@ export default class MultiLinePrompt extends Prompt<string> {
 				if (this.#showSubmit && this.focused === 'submit') {
 					this.focused = 'editor';
 				}
-				this.#insertAtCursor(char ?? '');
+				const casedChar = key.shift ? char.toUpperCase() : char;
+				this.#insertAtCursor(casedChar ?? '');
```
dreyfus92: i think this might break a couple of real cases:

- **paste**: terminals usually send pasted text as a stream of keypresses with `shift: false` per char. since the base `Prompt.onKeypress` already lowercased char, we'd never re-uppercase here, pasted mixed case ends up all lowercase.
- **capslock without shift**: some emulators send the upper-cased char with `shift: false` too, same deal.
the real root cause is that `Prompt.onKeypress` lowercases char before emitting 'key'. 

probably cleaner to just insert from `key.sequence` (the raw bytes, original case) and skip the `shift` dance entirely

wdyt @43081j ? 👀

MattStypa: I based this on existing code here: https://github.com/bombshell-dev/clack/blob/54be8d7a75c952b2157e94a78f70d40fa2d5321e/packages/core/src/prompts/select-key.ts#L26

But I see how that use-case is different. 

Let me take a closer look.

MattStypa:
@gameroman, @dreyfus92

Thank you for your feedback. The PR has been updated.

Please let me know if changing how prompt.ts emits key events is not the right solution in your opinion.

43081j:
i'm not sure about this as it touches more than it needs to.

cursor events purposely can be any `Action`, arrow keys don't hold any meaning here other than being some defaults. i.e. we don't need a concept of "arrow keys" or a new settings entry, as they are indeed just "actions".

i think the only changes you need are:

1. the key handler should pass `char` as-is (no lowercasing)
2. select-key refactor to account for this
3. update select-key test since it had the wrong casing already

you have all these so i think it looks good if we drop the arrow key stuff

gameroman:
> cursor events purposely can be any `Action`, arrow keys don't hold any meaning here other than being some defaults. i.e. we don't need a concept of "arrow keys" or a new settings entry, as they are indeed just "actions".

Do you mean they don't need to be in the settings file or don't need special handling?

Looking at the code, `CursorAction` is probably a better name for them instead of `ArrowKeyAction`

43081j:
im saying none of the changes around "arrow keys" are needed to fix this bug

MattStypa:
> im saying none of the changes around "arrow keys" are needed to fix this bug

@43081j  Im curious what your suggestion to fix the space key issue is.

At the moment, the multi-line key handler checks keys against `settings.actions` (which includes space) and returns early.

[see here](https://github.com/MattStypa/clack/blob/ab58ce54ce6c2d77528e0f024d76109fcc236c9a/packages/core/src/prompts/multi-line.ts#L92)

There are a many valid ways to address this. Initially, I scoped the set of cursor keys to `multi-line.ts` ([see here](https://github.com/bombshell-dev/clack/pull/534/changes/73b9c18244d9ac682fc1ced83fe8696186fde7ad#diff-5ae0fc8baba00efc9fc8e531cd9fd6b2316d99c9942a384f2b813b0966ad0567)).

43081j:
i think that is a bug in the multi-line prompt.

https://github.com/bombshell-dev/clack/blob/54be8d7a75c952b2157e94a78f70d40fa2d5321e/packages/core/src/prompts/multi-line.ts#L92-L95

here we check that the key is an "action" key, and treat it as a cursor action if it is.

but the multi-line prompt doesn't control its cursor actions based on the `actions` keys. it is hard coded to use the arrow keys here:

https://github.com/bombshell-dev/clack/blob/54be8d7a75c952b2157e94a78f70d40fa2d5321e/packages/core/src/prompts/multi-line.ts#L47-L58

the settings can't override that.

so its basically a bug i think - that we shouldn't be checking `settings.actions`. we should just have a hard coded check for if it is an arrow key or not.

so you were on the right track, but not as a change to the core settings or types. this is just a separate, one-off hard coded set of keys a user can't configure (specific to the multi-line prompt).

MattStypa:
@43081j updated

43081j (approved):

PR #543, opened 2026-05-19
feat: standard schema for validation

florian-lefebvre:
## What does this PR do?

Allows using Standard Schema compatible libraries for `validate()`. Discussed on Discord.

Docs PR: https://github.com/bombshell-dev/docs/pull/42

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist


examples/basic/standard-schema-validation.ts
```diff
@@ -0,0 +1,35 @@
...
+	});
+
+	if (!isCancel(validName)) {
+		note(`Valid name: ${validName}`, 'Success');
+	}
+
+	await setTimeout(1000);
+}
+
+main().catch(console.error);
```
ghostdevv: top level await? :p

packages/core/src/utils/validation.ts
```diff
@@ -0,0 +1,22 @@
+import type { StandardSchemaV1 } from '@standard-schema/spec';
+
+export type Validate<TValue> =
```
ghostdevv, line 35: Could use a jsdoc comment

ghostdevv (commented):
the jsdoc comments above the `validate` options would need updating, as well as the docs https://github.com/bombshell-dev/docs/

examples/basic/standard-schema-validation.ts
gameroman: Probably fine since it’s just an example

ghostdevv: something about being the change you want to see in the world :innocent:

examples/basic/standard-schema-validation.ts
```diff
@@ -0,0 +1,35 @@
...
+import { type } from 'arktype';
+
+async function main() {
+	console.clear();
+
+	// Example demonstrating the issue with initial value validation
+	const name = await text({
+		message: 'Enter your name (letters only)',
+		initialValue: 'John123', // Invalid initial value with numbers
+		validate: type('string.alpha').describe('Name can only contain letters'),
```
natemoo-re: To @ghostdevv s "be the change" comment, can we use a contrived number example or something neutral like a url? Names categorically [cannot be validated](https://usemy.name/principles/do-not-validate-names).

packages/core/src/utils/validation.ts
```diff
@@ -0,0 +1,22 @@
...
+export function runValidation<TValue>(
+	validate: Validate<TValue>,
+	value: TValue | undefined
+): string | Error | undefined {
+	if ('~standard' in validate) {
+		const result = validate['~standard'].validate(value);
+		// https://standardschema.dev/schema#how-to-only-allow-synchronous-validation
+		if (result instanceof Promise) {
+			throw new TypeError(
+				'Schema validation must be synchronous. Update `validate()` and get rid of any asynchronous logic.'
```
natemoo-re: Bummer that we're stuck with this! Maybe a follow-up for v2.

natemoo-re (commented):
Code looks great, changeset is perfect, just minor feedback on the example!

examples/basic/standard-schema-validation.ts
florian-lefebvre: FYI I just matched the text validation example but I will update it!

florian-lefebvre:
I also want to note that the PR does not support schema transforms (they will run but not be used). That would require more changes to represent input and output in types. Out of scope of this PR but if you're interested let me know!

florian-lefebvre:
@ghostdevv docs PR is available at https://github.com/bombshell-dev/docs/pull/42

packages/core/src/utils/validation.ts
florian-lefebvre: Added a todo comment

ghostdevv:
Also, this would be breaking right?

packages/core/src/utils/validation.ts
```diff
@@ -0,0 +1,35 @@
...
+	validate: Validate<TValue>,
+	value: TValue | undefined
+): string | Error | undefined {
+	if ('~standard' in validate) {
+		const result = validate['~standard'].validate(value);
+		// https://standardschema.dev/schema#how-to-only-allow-synchronous-validation
+		// TODO: https://github.com/bombshell-dev/clack/issues/92
+		if (result instanceof Promise) {
+			throw new TypeError(
+				'Schema validation must be synchronous. Update `validate()` and get rid of any asynchronous logic.'
```
43081j:
```suggestion
				'Schema validation must be synchronous. Update `validate()` and remove any asynchronous logic.'
```

packages/core/src/utils/validation.ts
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Represents the `validate()` option. A function or a
+ * [Standard Schema](https://github.com/standard-schema/standard-schema)
+ * that validates user input. Return a `string` or `Error` to show as a
+ * validation error, or `undefined` to accept the result.
+ */
+export type Validate<TValue> =
+	| ((value: TValue | undefined) => string | Error | undefined)
+	| StandardSchemaV1<TValue | undefined, any>;
```
43081j:
```suggestion
	| StandardSchemaV1<TValue | undefined, unknown>;
```

new code shouldn't really contain `any`. we should strongly type this

packages/core/package.json
```diff
@@ -56,10 +56,12 @@
     "test": "vitest run"
   },
   "dependencies": {
+    "@standard-schema/spec": "^1.1.0",
```
43081j: i'd rather not introduce an extra dependency for what is an optional feature (and a rarely used one for some time yet).

it seems we can either:

1. move it to `devDependencies` and hope the end-user installs it themselves (or skipLibCheck it)
2. don't use the spec DTS, just loosely accept something that has the right shape (i.e. ship our own much slimmer type)

florian-lefebvre:
@43081j yeah I was also hesitating. I inlined it, the spec says it's absolutely fine

florian-lefebvre:
> Also, this would be breaking right?

@ghostdevv by "this", are you referring to input/output/transforms or this PR?

ghostdevv:
> > Also, this would be breaking right?
> 
> @ghostdevv by "this", are you referring to input/output/transforms or this PR?

Yea, as this PR changes the behavior of the `validate` option

florian-lefebvre:
I'm not sure it's fully breaking. No change as a user but it's true that if you are dealing with lower level APIs like in the date prompt, then it requires changes. So maybe it's a major for core and a minor for prompts

ghostdevv:
> I'm not sure it's fully breaking. No change as a user but it's true that if you are dealing with lower level APIs like in the date prompt, then it requires changes. So maybe it's a major for core and a minor for prompts

Ah, I'm so sorry but I missed the fact that the old behavior is still available :woman_facepalming:

43081j:
technically, we can strip the standard schema types down to only the parts we use too. but i'm happy either way.

inlining it is better than pulling a dependency still

florian-lefebvre:
I just checked and very little can be stripped, only `InferInput` and `InferOutput`. So I'd say it's worth leaving it as is

43081j (approved):

examples/basic/standard-schema-validation.ts
```diff
@@ -0,0 +1,35 @@
+import { setTimeout } from 'node:timers/promises';
+import { isCancel, note, text } from '@clack/prompts';
+import { type } from 'arktype';
+
+async function main() {
```
ghostdevv: btw @florian-lefebvre what I meant by top level await was removing the `main` fn and just doing everything at the top - but it's not that deep 😄

packages/core/src/prompts/prompt.ts
```diff
@@ -13,12 +13,20 @@ import {
...
 export interface PromptOptions<TValue, Self extends Prompt<TValue>> {
 	render(this: Omit<Self, 'prompt'>): string | undefined;
 	initialValue?: any;
 	initialUserInput?: string;
-	validate?: ((value: TValue | undefined) => string | Error | undefined) | undefined;
+
+	/**
+	 * A function or a [Standard Schema](https://github.com/standard-schema/standard-schema)
+	 * that validates user input. Return a `string` or `Error` to show as a validation error,
+	 * or `undefined` to accept the result.
```
ghostdevv:
```suggestion
	 * A function or a [Standard Schema](https://github.com/standard-schema/standard-schema)
	 * that validates user input. If a custom function is given, you should return a `string` or `Error`
	 * to show as a validation error, or `undefined` to accept the result.
```

at the risk of bikeshedding, this is a bit more clear?

packages/core/src/utils/validation.ts
```diff
@@ -0,0 +1,35 @@
...
+ */
+export type Validate<TValue> =
+	| ((value: TValue | undefined) => string | Error | undefined)
+	| StandardSchemaV1<TValue | undefined, unknown>;
+
+/**
+ * Runs the `validate()` option and normalizes the result
+ * @param validate - The validate option
+ * @param value - The user input
+ * @returns string | Error | undefined
```
ghostdevv:
```suggestion
 * @returns the validation result
```

ghostdevv (changes requested):

packages/core/src/utils/validation.ts
```diff
@@ -0,0 +1,35 @@
+import type { StandardSchemaV1 } from './standard-schema.js';
+
+/**
+ * Represents the `validate()` option. A function or a
+ * [Standard Schema](https://github.com/standard-schema/standard-schema)
+ * that validates user input. If a custom function is given, you should return a
+ * `string` or `Error` to show as a validation error, or `undefined` to accept the result.
+ */
```
ghostdevv, line 34:
```suggestion
/**
 * A function or [Standard Schema](https://github.com/standard-schema/standard-schema)
 * that validates user input. If a custom function is given, you should return a
 * `string` or `Error` to show as a validation error, or `undefined` to accept the result.
 * 
 * @example Using arktype
 * ```ts
 * import { text } from '@clack/prompts';
 * import { type } from 'arktype';
 * 
 * const name = await text({
 *   message: 'Enter your name (letters only)',
 *   validate: type('string.alpha').describe('Name can only contain letters'),
 * });
 * ```
 *
 * @example Custom validator
 * ```ts
 * import { text } from '@clack/prompts';
 * 
 * const age = await text({
 *   message: 'Enter your age:',
 *   validate(value) {
 *     if (!value) return 'Please enter a value';
 *     const num = parseInt(value);
 *     if (isNaN(num)) return 'Please enter a valid number';
 *     if (num < 0 || num > 120) return 'Age must be between 0 and 120';
 *     return undefined;
 *   },
 * });
 * ```
 */
```

ghostdevv (changes requested):
last thing 😄 otherwise lgtm

dreyfus92 (approved):
thank you @florian-lefebvre 💜

PR #544, opened 2026-05-26
feat(prompts): add onCancel callback option to all prompts

### LINKED ISSUE #83: [Request] Improve cancellation API
We've gotten feedback that the `Symbol('clack:cancel')` approach _and_ the `onCancel` don't feel like the ideal API for handling cancellation, and I agree.

I'm opening this issue to continue the discussion about what an API for graceful cancellation could look like.

Seems like `AbortController` is designed specifically for this use case, so maybe there's some inspiration to take from the `signal` pattern?

Open to discussion!

**Additional context**
See https://github.com/natemoo-re/clack/pull/68

aqeelat:
## Summary

Adds an optional `onCancel` callback to all prompt functions via `CommonOptions`. When the user cancels a prompt (Ctrl+C or Escape), the callback is invoked — eliminating the need for `isCancel` guards at every call site. **When `onCancel` is provided, the return type narrows from `Promise<T | symbol>` to `Promise<T>`.**

Closes #83

## Motivation

Every clack user repeats this pattern:

```ts
const result = await confirm({ message: "Continue?" });
if (isCancel(result)) {
  cancel("Operation cancelled.");
  process.exit(0);
}
```

This PR makes it:

```ts
const result = await confirm({
  message: "Continue?",
  onCancel: () => {
    cancel("Operation cancelled.");
    process.exit(0);
  },
});
// result is `boolean` — no isCancel guard needed
```

## Changes

- **`CommonOptions`** — added `onCancel?: () => void`
- **`handleCancel` utility** — exported from `@clack/prompts` for custom prompt implementations
- **All 12 prompt functions** — converted to function declarations with overloads for return type narrowing
- **Tests** — runtime tests for cancel behavior + compile-time type narrowing tests

## Type narrowing

Each prompt uses function overloads so TypeScript narrows the return type when `onCancel` is present:

```ts
// Without onCancel: result is boolean | symbol
const a = await confirm({ message: "Continue?" });
//    ^? boolean | symbol

// With onCancel: result is boolean (narrowed)
const b = await confirm({ message: "Continue?", onCancel: () => { process.exit(0); } });
//    ^? boolean
```

This works for all prompts: `text`, `confirm`, `select`, `multiselect`, `groupMultiselect`, `password`, `autocomplete`, `autocompleteMultiselect`, `selectKey`, `date`, `multiline`, `path`.

## Design decisions

- **Additive, no breaking changes** — `onCancel` is optional; existing code works unchanged
- **Function overloads for narrowing** — the idiomatic TypeScript pattern; converts `export const` arrows to `export function` declarations
- **`handleCancel` is exported** — useful for custom prompt implementations

## Test plan

- [x] All 727 existing + new tests pass
- [x] `pnpm run build` succeeds
- [x] `pnpm run types` succeeds
- [x] `pnpm run lint` succeeds
- [x] `pnpm run format` succeeds
- [x] Type narrowing verified via compile-time assertions (`@ts-expect-error`)

### DIFF
--- a/.changeset/on-cancel-callback.md
+++ b/.changeset/on-cancel-callback.md
@@ -1,0 +1,24 @@
+---
+"@clack/prompts": patch
+---
+
+Add `onCancel` callback option to all prompts
+
+All prompt functions now accept an optional `onCancel` callback that is
+invoked when the user cancels (Ctrl+C or Escape). When the callback's
+return type is `never` (e.g. it calls `process.exit` or throws), the
+prompt's return type narrows to exclude the cancel symbol:
+
+```ts
+const result = await confirm({
+  message: 'Continue?',
+  onCancel: () => {
+    cancel('Operation cancelled.');
+    process.exit(0);
+  },
+});
+// result is `boolean` — no `isCancel` guard needed
+```
+
+For callbacks that return `void`, the return type remains `T | symbol` as
+before. Also exports `handleCancel` utility for custom prompt implementations.
--- a/packages/prompts/README.md
+++ b/packages/prompts/README.md
@@ -44,6 +44,33 @@ if (isCancel(value)) {
 }
 ```
 
+You can also provide an `onCancel` callback on any prompt to handle cancellation inline.
+When the callback calls `process.exit` or throws, the return type narrows to exclude the cancel symbol:
+
+```js
+import { cancel, text } from '@clack/prompts';
+
+const value = await text({
+  message: 'What is the meaning of life?',
+  onCancel: () => {
+    cancel('Operation cancelled.');
+    process.exit(0);
+  },
+});
+// value is `string` — no `isCancel` guard needed
+```
+
+Note: type narrowing only works when options are passed inline. If you extract options into a typed variable, the callback type widens to `() => void` and narrowing is lost:
+
+```js
+// Narrowing does NOT work here — result is `string | symbol`
+const opts: TextOptions = {
+  message: 'What is the meaning of life?',
+  onCancel: () => { process.exit(0); },
+};
+const value = await text(opts);
+```
+
 ## Components
 
 ### Text
--- a/packages/prompts/src/autocomplete.ts
+++ b/packages/prompts/src/autocomplete.ts
@@ -3,6 +3,7 @@ import type { Validate } from '@clack/core';
 import { AutocompletePrompt, settings } from '@clack/core';
 import {
 	type CommonOptions,
+	handleCancel,
 	S_BAR,
 	S_BAR_END,
 	S_CHECKBOX_INACTIVE,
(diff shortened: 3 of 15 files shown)


43081j:
> When onCancel is provided, the return type narrows from Promise<T | symbol> to Promise<T>.

Unsafely, yes:

```ts
const result = await confirm({
  message: "Continue?",
  onCancel: () => {
    console.log("do anything here that _doesn't_ process.exit");
  },
});

// result is still T | Symbol at runtime
```

i'm not sure we should go down this path yet until the issue has a resolution.

my main concern is that this is mixing two different architectures:

1. cancellations as symbols
2. cancellations as callbacks

we currently use the first one. introducing the second means we now have two ways of dealing with cancellations, which seems awkward to me.

ultimately you're just trying to avoid having to type-check results, but that takes about as much code as your `onCancel` handler. could you explain what the gain is?

aqeelat:
@43081j Thanks for the feedback! You were right that the previous version was unsound. I've updated the PR to address your concerns.

**Type narrowing is now sound** using `() => never` overloads:

```ts
export function confirm(opts: ConfirmOptions & { onCancel: () => never }): Promise<boolean>;
export function confirm(opts: ConfirmOptions): Promise<boolean | symbol>;
```

Narrowing only triggers when the callback's return type is `never` (e.g. `process.exit(0)` or `throw`). A callback like `() => console.log("cancelled")` returns `void`, so the return type stays `T | symbol` — no false narrowing. This is sound because TypeScript knows the callback never returns, making the cancel symbol unreachable.

**Symbols remain the single source of truth.** `onCancel` is just a convenience hook for the common pattern of log + exit. The callback fires before the symbol is returned; if it throws, the promise rejects and the symbol is never returned.

**The gain isn't about saving lines** — it's about the compiler enforcing that cancellation is handled. With `isCancel`, nothing prevents a caller from forgetting the guard, leading to `symbol` values leaking into business logic (only caught at runtime). With `() => never`, the compiler proves the cancel path is handled.

43081j:
> With isCancel, nothing prevents a caller from forgetting the guard, leading to symbol values leaking into business logic (only caught at runtime).

are you explaining your changes, or are you explaining the current isCancel?

the current isCancel in main does indeed get caught at _build time_ (because the return type is `T | Symbol`).

what i'm asking for is why this change is useful _at all_. we need some context here. what can you do now that you couldn't do before with an isCancel guard?

ghostdevv:
It certainly feels like supporting `onCancel` means that you're just shifting the burden up a bit, but I assume with the benefit that you could do something like

```ts
import { confirm } from '@clack/prompts';
import { onCancel } from './shared.ts';

const confirmation = await confirm({ message: 'you sure?', onCancel });
//        ^? boolean
```

However, you can achieve a similar result using an assertion type guard

```ts
import { confirm, isCancel, cancel } from '@clack/prompts';
import process from 'node:process';

function handleCancel<T>(result: T): asserts result is Exclude<T, symbol> {
    if (isCancel(result)) {
        cancel('Cancelled');
        process.exit(1);
    }
}

const confirmation = await confirm({ message: 'you sure?' });
handleCancel(confirmation);

confirmation;
// ^? boolean
```

If this approach works for you that'd be ideal, because as James said otherwise there would be two ways of dealing with the same problem and it kinda adds confusion and messes with the control flow in clack.

Oh and the other option is of course [using a group](https://bomb.sh/docs/clack/packages/prompts/#group), but obviously may not make sense in all situations.

43081j:
going to close this one. we can discuss in future if there's demand for it but the existing `isCancel` seems fine for now and this design would just cause duplication, so needs more thought.

PR #559, opened 2026-06-09
feat(core): add ctrl+n/ctrl+p keybindings for navigation

rafa-thayto:
## Summary

Adds emacs-style `ctrl+n` / `ctrl+p` keybindings as default aliases for `down` / `up` navigation, alongside the existing vim `j` / `k` aliases.

- **`settings.ts`** — registers `'\x10'` → `up` (ctrl+p) and `'\x0e'` → `down` (ctrl+n) in the default alias table under an `// emacs support` block; adds a `getActionForControlKey(char, key)` helper (sibling of `isActionKey`) that resolves control chords, guarded on `key.ctrl` so typed letters can never alias navigation in text inputs.
- **`prompt.ts`** — the cursor-alias lookup in `onKeypress` now checks `[char, key.name, key.sequence]` instead of `key.name` only, mirroring how the pre-existing `'\x03'` (ctrl+c) cancel alias is matched. Control chords report `key.name` as the bare letter (`'n'`), so the raw sequence is the only safe key to match on.
- **`autocomplete.ts`** — `ctrl+n`/`ctrl+p` navigate the option list (the prompt tracks text input, so it reads `key.name === 'up'/'down'` directly and needed explicit chord resolution).
- **`multi-line.ts`** — `ctrl+n`/`ctrl+p` move the cursor between lines; previously the raw `\x0e`/`\x10` bytes were inserted into the user's text.

Every cursor-driven prompt (`select`, `multi-select`, `group-multiselect`, `date`, `confirm`) picks the aliases up automatically through the existing cursor-event system, and `updateSettings({ aliases })` keeps working unchanged for user-defined sequences.

## Test plan

- New unit tests: cursor events emitted for ctrl+p/ctrl+n (and **not** for plain `n`/`p`), autocomplete list navigation, multi-line cursor movement without text mutation.
- New integration tests with snapshots: `select` responds to ctrl+n/ctrl+p.
- Full suite green: 144 core + 587 prompts tests; `biome check` and `tsc --noEmit` clean.
- Changeset included (`@clack/core` minor).

### DIFF
--- a/.changeset/ctrl-n-ctrl-p-navigation.md
+++ b/.changeset/ctrl-n-ctrl-p-navigation.md
@@ -1,0 +1,5 @@
+---
+'@clack/core': minor
+---
+
+Add emacs-style `ctrl+n` and `ctrl+p` keybindings for `down`/`up` navigation. They work as default aliases in every cursor-driven prompt (`select`, `multi-select`, `group-multiselect`, `date`, `confirm`), navigate the option list in `autocomplete`, and move the cursor between lines in `multi-line` (where the raw control bytes were previously inserted into the text).
--- a/packages/core/src/prompts/autocomplete.ts
+++ b/packages/core/src/prompts/autocomplete.ts
@@ -1,6 +1,7 @@
 import type { Key } from 'node:readline';
 import { styleText } from 'node:util';
 import { findCursor } from '../utils/cursor.js';
+import { getActionForControlKey } from '../utils/index.js';
 import Prompt, { type PromptOptions } from './prompt.js';
 
 interface OptionLike {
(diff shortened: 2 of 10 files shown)


rafa-thayto:
I was implementing this on my CLI, but I thought it would probably be a better idea to implement it directly here!

rafa-thayto:
@dreyfus92

dreyfus92 (commented):
hey @rafa-thayto, thanks for putting this together, the write-up is super clear and clean. but i don't think we're going to merge this one. clack already ships arrow keys + vim(hjkl) for navigation, and we'd rather keep the default keymap lean than grow a third set of bindings into core. it's less about this change specifically and more about where we draw the line on what's built in by default.

we're also starting to ask folks to open an issue before a pr so we align on direction first, no worries on this one since that wasn't in place yet.

we really appreciate you taking a swing at this 😄

PR #567, opened 2026-06-16
feat(prompts): add instructions footer for `select`, `multi-select` & `group-multi-select` prompts

### LINKED ISSUE #558: [Request] Show instructions for `multiselect` and `groupMultiselect`
We had reports that it's unclear how to select items (https://github.com/changesets/changesets/issues/255)

It'd be nice to show instructions for selecting for `multiselect` and `groupMultiselect`, similar to `autocomplete` and `autocompleteMultiselect` today:

Image

Maybe this could be under an option whether to show by default.

https://github.com/bombshell-dev/clack/issues/551 is a similar issue of a more general footer option instead of like an `instructions: boolean` option

**Additional context**
n/a

dreyfus92:
## What does this PR do?

adds `instructions` to `select`, `multi-select` and `group-multi-select` that _always_ shows a persistent keyboard hints footer in the active state, matching autocomplete styling, with tests. 

Closes #558

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist


packages/prompts/src/common.ts
```diff
@@ -73,3 +73,21 @@ export interface CommonOptions {
 	signal?: AbortSignal;
 	withGuide?: boolean;
 }
+
+export const MULTISELECT_INSTRUCTIONS = [
```
43081j: i think this should actually live in `select.ts` - which group-multi-select and multi-select already depend on.

then we can introduce the option to all 3

packages/prompts/src/common.ts
```diff
@@ -73,3 +73,21 @@ export interface CommonOptions {
...
+	`${styleText('dim', '↑/↓')} to navigate`,
+	`${styleText('dim', 'Space:')} select`,
+	`${styleText('dim', 'Enter:')} confirm`,
+];
+
+export function formatInstructionFooter(
+	instructions: string[] | null,
+	hasGuide: boolean
+): { text: string; lineCount: number } {
```
43081j: i think it should just return `string[]` and the consumer can do the joining/counting

packages/prompts/src/common.ts
dreyfus92: should i also move the `formatInstructionFooter` utility along with this one? or leave it in `common.ts`

packages/prompts/src/group-multi-select.ts
```diff
@@ -285,9 +293,13 @@ export const groupMultiselect = <Value>(opts: GroupMultiSelectOptions<Value>) =>
 				}
 				default: {
 					const guidePrefix = hasGuide ? `${styleText('cyan', S_BAR)}  ` : '';
-					// Calculate rowPadding: title lines + footer lines (S_BAR_END + trailing newline)
 					const titleLineCount = title.split('\n').length;
-					const footerLineCount = (hasGuide ? 1 : 0) + 1; // guide line + trailing newline
+					const footerLines = formatInstructionFooter(
+						showInstructions ? MULTISELECT_INSTRUCTIONS : null,
```
43081j: if `!showInstructions` this is no-op right? so should probably be something more like `showInstructions ? formatInstructionFooter(...) : []`

packages/prompts/src/select.ts
```diff
@@ -2,15 +2,26 @@ import { styleText } from 'node:util';
...
 	symbolBar,
 } from './common.js';
 import { limitOptions } from './limit-options.js';
 
+export const SELECT_INSTRUCTIONS = [
+	`${styleText('dim', '↑/↓')} to navigate`,
+	`${styleText('dim', 'Enter:')} confirm`,
+];
+
+export const MULTISELECT_INSTRUCTIONS = [
```
43081j: these belong in multi-select i guess, which group-multi-select can depend on

dreyfus92: i moved them to `select.ts` as you requested, but  i can move them to the multiselect prompt tho.

packages/prompts/src/group-multi-select.ts
dreyfus92: ohh, nice catch! null isn't a no-op; with `hasGuide` it still returns [`S_BAR_END`].

so it should be something like

```ts
showInstructions ? formatInstructionFooter(...) : hasGuide ? [S_BAR_END] : [] 
```

packages/prompts/src/select.ts
43081j: that was back when we had only one 😄 

they belong alongside what they're named after:

- `MULTISELECT_INSTRUCTIONS` -> multi-select
- `SELECT_INSTRUCTIONS` -> select

import elsewhere

packages/prompts/src/common.ts
43081j: we may have other footers in future so its probably fine

packages/prompts/src/group-multi-select.ts
43081j: i think so yes, good catch 👍  at that point lets do it as an `if` though rather than nested ternaries

43081j (approved):

43081j (approved):

menukaDad (commented):
مستوره الواطيه

menukaDad (approved):

PR #569, opened 2026-06-16
fix(multi-line): only submit on double-return at end of value

43081j:
## What does this PR do?

Closes #566

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor (no behavior change)
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests
- [ ] Chore (dependencies, CI, tooling)

## Checklist

- [x] `pnpm test` passes (or targeted tests for my change)
- [x] `pnpm format` has been run
- [x] I have added/updated tests for my changes (if applicable)
- [x] I have added a changeset

## AI-generated code disclosure

- [ ] This PR includes AI-generated code

PR #570, opened 2026-06-16
fix(core): seed multiline prompt with initialValue

### LINKED ISSUE #566: [Bug] `initialValue` on `multiline`not working
**Environment**

- OS: macOS
- Node Version: v26.3.0
- Package: `@clack/prompts`
- Package Version: 1.5.1

**Describe the bug**
Passing `initialValue` to `multiline` doesn't have any effect.

**To Reproduce**
https://stackblitz.com/edit/node-4yxn6jvr

- Run `node index.js` in the terminal. 

**Expected behavior**
A multiline input with the initial value in it.

**Additional Information**

Image

spokodev:
Closes #566.

`multiline` accepts `initialValue` but never seeded the editor with it, so passing it had no effect — the prompt always opened empty.

`TextPrompt` forwards `initialValue` to the base `Prompt` as `initialUserInput`; `MultiLinePrompt` did not. This mirrors that single line in the `MultiLinePrompt` constructor, so the editor opens pre-filled and an unedited submit returns the seeded value (falling back to `initialUserInput`).

### Tests
Two `@clack/core` tests, red before the change and green after:
- seeds `initialValue` into the editor (`userInput` + `value`)
- submits `initialValue` when not edited

Full `@clack/core` suite: 143 passing. Changeset included (`@clack/core` patch).

### DIFF
--- a/.changeset/multiline-initial-value.md
+++ b/.changeset/multiline-initial-value.md
@@ -1,0 +1,5 @@
+---
+"@clack/core": patch
+---
+
+Fix `multiline` prompt ignoring `initialValue`. The editor is now seeded with `initialValue` (falling back to `initialUserInput`) when the prompt opens, matching the `text` prompt.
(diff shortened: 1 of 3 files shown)


43081j:
thanks for the contribution 🙏  but this is already being fixed in the earlier #569 PR

PR #574, opened 2026-06-22
feat(prompts): add showInstructions opt-out

### LINKED ISSUE #571: [Request] Allow disabling keyboard hints via options for select-like components

Refs: 
* #567

Recently a permanent keyboard hints has been added to select/multiselect/group multiselect components.

Add option to all components listed above to disable (hide) the hint. The new option can be `false` by default (hint shown by default, to match the latest release).

Do not update `@clack/prompts` to the latest release.

**Additional context**

It's a nice feature, but there are cases when I want to prevent the hints from showing, also increased degree of customization would be an improvement anyway.

dreyfus92:
## What does this PR do?

adds `showInstructions` opt-out to `select`, `multiselect` and `group-multiselect` prompts

Closes #571

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

### DIFF
--- a/.changeset/fresh-facts-search.md
+++ b/.changeset/fresh-facts-search.md
@@ -1,0 +1,5 @@
+---
+"@clack/prompts": minor
+---
+
+Add `showInstructions` option to `select`, `multiselect`, and `groupMultiselect`. Keyboard hints remain shown by default; pass `showInstructions: false` to hide them.
--- a/packages/prompts/src/group-multi-select.ts
+++ b/packages/prompts/src/group-multi-select.ts
@@ -60,6 +60,12 @@ export interface GroupMultiSelectOptions<Value> extends CommonOptions {
 	 * @default 0
 	 */
 	groupSpacing?: number;
+
+	/**
+	 * Show keyboard instructions below the option list.
+	 * @default true
+	 */
+	showInstructions?: boolean;
 }
 
 /**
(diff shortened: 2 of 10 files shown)


.changeset/fresh-facts-search.md
```diff
@@ -0,0 +1,5 @@
+---
+"@clack/prompts": minor
+---
+
+Add `showInstructions` option to `select`, `multiselect`, and `groupMultiselect`. Keyboard hints remain shown by default; pass `showInstructions: false` to hide them.
```
natemoo-re, line 5: In favor of the toggle, but wondering if we should set the default to `false` to maintain back-compat and flip the default in v2? That would make this a `patch` change.

dreyfus92: i happy either way, let's see what james thinks about this @43081j 👀

natemoo-re: I know the logic was that some prompts had instructions and some didn't, so this was framed as a standardization, which means we probably can't avoid a default behavior change in some cases. Either way we go, providing an option is the right move.

43081j: this was already released afaik.

so i wouldn't bother going back on ourselves and would just keep defaulted to `true`

43081j (approved):

.changeset/fresh-facts-search.md
natemoo-re: Works for me!

natemoo-re (approved):

dreyfus92:
thank you both! 💜

PR #576, opened 2026-06-24
feat(core,prompts): export CANCEL_SYMBOL

### LINKED ISSUE #554: Export the cancel symbol
## Context

This came up while evaluating `@clack/prompts` as a potential replacement for `enquirer.js`, comparing it with `enquirer.js` and `@inquirer/prompts`.

Split out from: #550

## Problem

`isCancel(value)` identifies cancellation values, but the actual cancel symbol is not exported.

This makes a few things harder:

- Tests cannot easily create a cancellation value that behaves like clack's own cancel result
- Prompt return types use the broad `symbol` type
- `isCancel` narrows to `symbol` rather than a specific cancel symbol type

## Proposal

Export the cancel symbol and a corresponding type from `@clack/core` and `@clack/prompts`:

```ts
import { CANCEL_SYMBOL, isCancel } from '@clack/prompts'
import type { CancelSymbol } from '@clack/prompts'
```

Prompt return types could then use `CancelSymbol` instead of `symbol`:

```ts
Promise<string | CancelSymbol>
```

This would allow more precise narrowing:

```ts
const result = await text({ message: 'Name?' })

if (isCancel(result)) {
  // result: CancelSymbol
} else {
  // result: string
}
```

luisangelrod:
## What does this PR do?

Exports `CANCEL_SYMBOL` from `@clack/core` and `@clack/prompts` so consumers can reference the symbol directly without reimplementing it.

Closes #554

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [ ] `pnpm test` passes (or targeted tests for my change)
- [ ] `pnpm format` has been run


43081j:
please follow the PR template provided rather than whatever your agent output:
https://github.com/bombshell-dev/clack/blob/main/.github/PULL_REQUEST_TEMPLATE.md

i also don't think we need to export the type as it can very easily be inferred via `typeof`. We were just missing the export of `CANCEL_SYMBOL`.

luisangelrod:
Thanks for the feedback. Updated the PR description to follow the template and simplified the change to just export CANCEL_SYMBOL without the explicit type.

.changeset/bright-wolves-export-cancel.md
```diff
@@ -0,0 +1,9 @@
+---
+"@clack/core": minor
+"@clack/prompts": minor
+---
+
+Export `CANCEL_SYMBOL` constant and `CancelSymbol` type from both packages.
```
43081j, line 6: you can discard this changeset now

.changeset/shaggy-boxes-crash.md
```diff
@@ -0,0 +1,6 @@
+---
+"@clack/core": minor
+"@clack/prompts": minor
+---
+
+Export `CANCEL_SYMBOL` constant and `CancelSymbol` type from `@clack/core` and `@clack/prompts`. Update `isCancel` type guard to narrow to `CancelSymbol` instead of the broad `symbol` type.
```
43081j, line 6: and this one no longer needs to mention the type

oaksprout:
We can run this PR through an independent check if you want. The diff is applied to a clean copy at its base commit, your test suite and typechecks run in a locked-down container, and a separate pass reviews the change against #554. Results come back as a comment with a record you can inspect. Free, no obligation, and we don't touch the PR itself.

AI agents do the checking; a human reviews everything before it's posted and answers replies. If that's useful, say so. If not, ignore this and we won't post again.

PR #578, opened 2026-06-30
fix(prompts): truncate wide-character box titles by display width

mahirhir:
## What does this PR do?

`box()` measures the title and every line with `fast-string-width`, but it truncates an over-long title with `title.slice(0, maxTitleLength - 3)`, which counts UTF-16 code units rather than display columns. `maxTitleLength` is a display-width budget, so for a wide-character (CJK/full-width) title the sliced result is still far too wide. The top border comes out ragged, and once `getPaddingForLine` returns a negative remainder, `'─'.repeat(...)` throws `RangeError: Invalid count value`.

Repro on `main`:

```ts
import { box } from '@clack/prompts';

// '这是一个非常长的标题' is 10 full-width chars = 20 display columns
box('message', '这是一个非常长的标题', { width: 0.2 });
// RangeError: Invalid count value: -6
```

The fix replaces the code-unit slice with a small `truncateToWidth` helper that accumulates characters by display width, the same way the rest of `box()` already measures. For ASCII titles it returns exactly what the old slice did, so existing snapshots are unchanged; only wide-character titles change behaviour (they now fit the box and the borders stay aligned).

No linked issue: this is a small self-contained bug fix.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

### DIFF
--- a/packages/prompts/src/box.ts
+++ b/packages/prompts/src/box.ts
@@ -96,6 +96,22 @@ function getPaddingForLine(
 	return [leftPadding, rightPadding];
 }
 
+function truncateToWidth(str: string, maxWidth: number): string {
+	const ellipsis = '...';
+	const target = maxWidth - stringWidth(ellipsis);
+	let result = '';
+	let width = 0;
+	for (const char of str) {
+		const charWidth = stringWidth(char);
+		if (width + charWidth > target) {
+			break;
+		}
+		result += char;
+		width += charWidth;
+	}
+	return result + ellipsis;
+}
+
 const defaultFormatBorder = (text: string) => text;
 
 /**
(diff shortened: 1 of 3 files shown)


packages/prompts/src/box.ts
```diff
@@ -96,6 +96,22 @@ function getPaddingForLine(
 	return [leftPadding, rightPadding];
 }
 
+function truncateToWidth(str: string, maxWidth: number): string {
+	const ellipsis = '...';
+	const target = maxWidth - stringWidth(ellipsis);
```
43081j, line 101: the `ellipsis` is constant so should be moved out of the function, and the length doesn't need to use `stringWidth` since we know it has no wide characters

packages/prompts/src/box.ts
```diff
@@ -96,6 +96,22 @@ function getPaddingForLine(
 	return [leftPadding, rightPadding];
 }
 
+function truncateToWidth(str: string, maxWidth: number): string {
```
43081j, line 99: this should really be in `common.ts` since it will be useful in other places too

packages/prompts/src/box.ts
```diff
@@ -96,6 +96,22 @@ function getPaddingForLine(
...
+	const target = maxWidth - stringWidth(ellipsis);
+	let result = '';
+	let width = 0;
+	for (const char of str) {
+		const charWidth = stringWidth(char);
+		if (width + charWidth > target) {
+			break;
+		}
+		result += char;
+		width += charWidth;
```
43081j, line 110:
```suggestion
		width += charWidth;
		if (width > target) {
			break;
		}
		result += char;
```

to avoid computing it twice

PR #583, opened 2026-07-02
feat: add async validation

### LINKED ISSUE #92: Async validation

43081j:
This adds the ability to have async validation to all prompts.

**For now, only the text prompt will render a validating message.**

Every other prompt needs its own individual follow-up PR to add support for rendering the new `validating` state. They will all accept an async validator but will not show a message while validating right now.

Closes #92 

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

### DIFF
--- a/packages/core/src/prompts/prompt.ts
+++ b/packages/core/src/prompts/prompt.ts
@@ -210,6 +210,10 @@ export default class Prompt<TValue> {
 	}
 
 	private async onKeypress(char: string | undefined, key: Key) {
+		if (this.state === 'validating') {
+			return;
+		}
+
 		if (this._track && key.name !== 'return') {
 			if (key.name && this._isActionKey(char, key)) {
 				this.rl?.write(null, { ctrl: true, name: 'h' });
--- a/packages/core/src/prompts/prompt.ts
+++ b/packages/core/src/prompts/prompt.ts
@@ -209,7 +209,7 @@ export default class Prompt<TValue> {
 		this._setUserInput('');
 	}
 
-	private onKeypress(char: string | undefined, key: Key) {
+	private async onKeypress(char: string | undefined, key: Key) {
 		if (this._track && key.name !== 'return') {
 			if (key.name && this._isActionKey(char, key)) {
 				this.rl?.write(null, { ctrl: true, name: 'h' });
@@ -238,7 +238,21 @@ export default class Prompt<TValue> {
 
 		if (key?.name === 'return' && this._shouldSubmit(char, key)) {
 			if (this.opts.validate) {
-				const problem = runValidation(this.opts.validate, this.value);
+				const problemResult = runValidation(this.opts.validate, this.value);
+				let problem: string | Error | undefined;
+				// Only if it is not a string or an Error, we assume
+				// it is a Promise and await it.
+				if (
+					problemResult !== undefined &&
+					typeof problemResult !== 'string' &&
+					!(problemResult instanceof Error)
+				) {
+					this.state = 'validating';
+					this.render();
+					problem = await problemResult;
+				} else {
+					problem = problemResult;
+				}
 				if (problem) {
 					this.error = problem instanceof Error ? problem.message : problem;
 					this.state = 'error';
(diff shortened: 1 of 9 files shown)


packages/core/src/utils/validation.ts
```diff
@@ -45,15 +47,13 @@ export type Validate<TValue> =
 export function runValidation<TValue>(
 	validate: Validate<TValue>,
 	value: TValue | undefined
-): string | Error | undefined {
+): MaybePromise<string | Error | undefined> {
 	if ('~standard' in validate) {
 		const result = validate['~standard'].validate(value);
 		// https://standardschema.dev/schema#how-to-only-allow-synchronous-validation
 		// TODO: https://github.com/bombshell-dev/clack/issues/92
```
florian-lefebvre: Can be removed

packages/core/src/prompts/prompt.ts
```diff
@@ -209,7 +209,7 @@ export default class Prompt<TValue> {
 		this._setUserInput('');
 	}
 
-	private onKeypress(char: string | undefined, key: Key) {
+	private async onKeypress(char: string | undefined, key: Key) {
```
dreyfus92, line 227: this is now async but the emitter doesn't await it, so input keeps flowing while a validation is in flight. so i feel like there are 2 problems maybe i'm overthinking u tell me:
- press return twice fast and you get 2 concurrent validations. the first one can resolve, emit `submit`, close the prompt and then second resolves and tries to render/emit on a finalizad prompt. state gets tangled.
- the user can keep typing while validating, mutating `value` mid-flight. then the error or submit applies to a stale value, and the rendered dimmed input won't match what was validated.

packages/core/src/prompts/prompt.ts
```diff
@@ -238,7 +238,21 @@ export default class Prompt<TValue> {
...
+				// Only if it is not a string or an Error, we assume
+				// it is a Promise and await it.
+				if (
+					problemResult !== undefined &&
+					typeof problemResult !== 'string' &&
+					!(problemResult instanceof Error)
+				) {
+					this.state = 'validating';
+					this.render();
+					problem = await problemResult;
```
dreyfus92, line 265: if the async validator throws (network error, w/e), this will throw inside an un-awaited async handler. that's an unhandled promise rejection, which kills the process in node. probably it might be better to do a try/catch? so that routes the thrown error into `this.error` / `state = 'error'` instead. also maybe would be worth a test case too.

packages/core/src/prompts/prompt.ts
```diff
@@ -238,7 +238,21 @@ export default class Prompt<TValue> {
...
 			if (this.opts.validate) {
-				const problem = runValidation(this.opts.validate, this.value);
+				const problemResult = runValidation(this.opts.validate, this.value);
+				let problem: string | Error | undefined;
+				// Only if it is not a string or an Error, we assume
+				// it is a Promise and await it.
+				if (
+					problemResult !== undefined &&
+					typeof problemResult !== 'string' &&
+					!(problemResult instanceof Error)
```
dreyfus92:
```suggestion
					problemResult instanceof Promise
```

wouldn't be a more clean approach for the check to be set this way?

dreyfus92 (approved):
awesome job james, left a few comments 👀

packages/core/src/prompts/prompt.ts
43081j: good catch!

we may be able to just return early when we're validating, i.e. ignore all key presses during validation

packages/core/src/prompts/prompt.ts
43081j: this is already true for synchronous validators. today, if a sync validator throws, it will go uncaught as you're not meant to throw inside one.

so changing this isn't really related to this PR, but could be argued as improving the DX

maybe lets do it in a different PR?

dreyfus92: sounds good, but would be nice if you could add an issue so it doesn't rot in time

dreyfus92 (approved):
thank you @43081j ✌🏻

PR #584, opened 2026-07-10
fix: wrap and indent `spinner` messages

beeequeue:
## What does this PR do?

makes sure the printed messages in `spinner` and its dependent prompts are correctly wrapped and indented.

<details>
<summary>Screenshots</summary>

Before:

image

image

After:

image

image

</details>

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

---

i spotted #479 which might also fix this issue, but it seems to have lost momentum and i think this is a reasonable patch until it's done

### DIFF
--- a/packages/prompts/src/spinner.ts
+++ b/packages/prompts/src/spinner.ts
@@ -1,5 +1,5 @@
 import { styleText } from 'node:util';
-import { block, getColumns, settings } from '@clack/core';
+import { block, getColumns, settings, wrapTextWithPrefix } from '@clack/core';
 import { wrapAnsi } from 'fast-wrap-ansi';
 import { cursor, erase } from 'sisteransi';
 import {
@@ -127,15 +127,16 @@ export const spinner = ({
 		return min > 0 ? `[${min}m ${secs}s]` : `[${secs}s]`;
 	};
 
+	const prefix = `${styleText('gray', S_BAR)}  `;
 	const hasGuide = opts.withGuide ?? settings.withGuide;
 
 	const start = (msg = ''): void => {
 		isSpinnerActive = true;
 		unblock = block({ output });
-		_message = removeTrailingDots(msg);
+		_message = wrapTextWithPrefix(output, removeTrailingDots(msg), hasGuide ? prefix : '', '');
 		_origin = performance.now();
 		if (hasGuide) {
-			output.write(`${styleText('gray', S_BAR)}\n`);
+			output.write(`${prefix}\n`);
 		}
 		let frameIndex = 0;
 		let indicatorTimer = 0;
@@ -159,11 +160,7 @@ export const spinner = ({
 				outputMessage = `${frame}  ${_message}${loadingDots}`;
 			}
 
-			const wrapped = wrapAnsi(outputMessage, columns, {
-				hard: true,
-				trim: false,
-			});
-			output.write(wrapped);
+			output.write(outputMessage);
 
 			frameIndex = frameIndex + 1 < frames.length ? frameIndex + 1 : 0;
 			// indicator increase by 1 every 8 frames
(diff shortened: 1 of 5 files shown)


packages/prompts/src/spinner.ts
```diff
@@ -159,11 +160,7 @@ export const spinner = ({
 				outputMessage = `${frame}  ${_message}${loadingDots}`;
 			}
 
-			const wrapped = wrapAnsi(outputMessage, columns, {
-				hard: true,
-				trim: false,
-			});
-			output.write(wrapped);
+			output.write(outputMessage);
```
dreyfus92, line 163: looks like `message()` slipped thru, it still sets `_message` raw in line 203 but the per frame wrap that used to catch it is gone now. so anything long/multiline set mid-spin renders unwrapped until stop. should probable mirror what `start()` does on line 136 and also a test calling `.message()` with a `columns + 10` string would cover the gap.

packages/prompts/src/spinner.ts
```diff
@@ -127,15 +127,16 @@ export const spinner = ({
 		return min > 0 ? `[${min}m ${secs}s]` : `[${secs}s]`;
 	};
 
+	const prefix = `${styleText('gray', S_BAR)}  `;
```
dreyfus92, line 130: head up that `wrapTextWithPrefix` does `column - prefix.length` and this prefix has ansi codes in it, so it's ~13 chars instead of 3 visible cols. messages wrap ~ 10 columns early.  not this PR's fault necessarily but passing a styled prefix here surfaces it, might be worth stripping ansi (or using visible width) in core as follow up.

packages/prompts/src/spinner.ts
```diff
@@ -127,15 +127,16 @@ export const spinner = ({
...
 
 	const start = (msg = ''): void => {
 		isSpinnerActive = true;
 		unblock = block({ output });
-		_message = removeTrailingDots(msg);
+		_message = wrapTextWithPrefix(output, removeTrailingDots(msg), hasGuide ? prefix : '', '');
 		_origin = performance.now();
 		if (hasGuide) {
-			output.write(`${styleText('gray', S_BAR)}\n`);
+			output.write(`${prefix}\n`);
```
dreyfus92, line 139: could we keep this as just the bar? the 2 trailing spaces from prefix are why basically every snapshot in both files changed. writing trailing whitespace doesn't buy anything and would shrink the diff a ton.

dreyfus92 (commented):
awesome job @beeequeue left a few comments, it would be nice that @43081j also jumps into this ✌🏻

packages/prompts/src/spinner.ts
```diff
@@ -159,11 +160,7 @@ export const spinner = ({
 				outputMessage = `${frame}  ${_message}${loadingDots}`;
 			}
 
-			const wrapped = wrapAnsi(outputMessage, columns, {
-				hard: true,
-				trim: false,
-			});
-			output.write(wrapped);
```
dreyfus92, line 166: pre-existing rather than introduced here, but this diff changes its failure mode: `clearPrevMessage` counts erase lines from `_prevMessage` alone, but what's on screen is `${frame}  ${_message}${loadingDots}` which is 3 cols wider. with `withGuide: false` and a near-full-width message, the output soft-wraps one row more than counted and a stale line survives the erase. repro (fake 40-col output, `start('x'.repeat(40))`):

```sh
main:        "•  "                                        <- stale stub
             "o  done"
this branch: "•  xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"   <- stale full line
             "o  done"
```

so same bug, but without the write-time wrap the leftover is now a full line of content instead of a 3-char stub. proper fix is probably counting from the actually-written string (needs its own var, the `isCI` dedup compares `_message === _prevMessage` and the frame changes every tick). fine as a follow-up, just wanted it on record

43081j:
there's been a few attempts at this same change but i've always held off on it since the real fix is to make the spinner use the prompt API.

we're a bit blocked there though since we'd end up reintroducing the windows spinner bug (main has one too, just a different bug, including in this PR).

i think lets hold off on merging this until i can catch myself up on the state of the prompt PR. if we can land that anytime soon then it'd make this one redundant

PR #581, opened 2026-07-01
fix: run spinner onCancel on Ctrl+C

cyphercodes:
## What does this PR do?

Fixes spinner cancellation from <kbd>Ctrl</kbd>+<kbd>C</kbd> when the keypress is handled by the blocked input path. The spinner now treats the blocked-input exit path as cancellation, passes the configured input stream to `block()`, runs `onCancel`, and sets `isCancelled`.

Closes #573

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor (no behavior change)
- [ ] Documentation
- [ ] Performance improvement
- [x] Tests
- [ ] Chore (dependencies, CI, tooling)

## Checklist

- [x] `pnpm test` passes (or targeted tests for my change)
- [x] `pnpm format` has been run
- [x] I have added/updated tests for my changes (if applicable)
- [x] I have added a changeset

## AI-generated code disclosure

- [x] This PR includes AI-generated code

PR #587, opened 2026-07-24
[Bug] Prevent hard process exit on Ctrl+C if an onCancel callback is passed to spinner

### LINKED ISSUE #573: [Bug] Spinner does not run `onCancel` function on `Ctrl + C`
**Environment**

- OS: macOS 15.7.7
- Node Version: v26.1.0
- Package: `@clack/prompts`
- Package Version: v1.6.0

**Describe the bug**
When spinner is spinning, `Ctrl + C` finishes the spinner with `Cancelled` message, but does not execute `onCancel` function

**To Reproduce**
[spinner-cancel example](https://github.com/bombshell-dev/clack/blob/%40clack/prompts%401.6.0/examples/basic/spinner-cancel.ts) can be used as a reproduction

- Use `spinner-cancel` example
- Run the example
- Press `Ctrl + C`

**Expected behavior**

`onCancel` should run and print `note` log

**Additional Information**

https://github.com/user-attachments/assets/26209b73-ba12-4409-973c-1fbcc1369354

kajoseph:
## What does this PR do?

With the spinner running, `Ctrl+C` hard-exits the node process with a buried `process.exit(0)` as noted by @43081j [here](https://github.com/bombshell-dev/clack/pull/581#issuecomment-4864121771). This short-circuits the spinner stop cycle which prevents the `onCancel` callback from running and also potentially leaves dangling connections to other services (e.g. a database).

This PR changes the behavior slightly so that an `onCancel` callback will prevent the cancellation cycle from calling `process.exit()` and instead return control back to the program so that it can cleanly teardown connections or otherwise proceed as desired. This modified behavior is more in-line with the rest of the prompts in that a `Ctrl+C` returns control back to the program instead of treating it as a SIGINT.

Closes https://github.com/bombshell-dev/clack/issues/573

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [x] This PR includes AI-generated code*

\* only the added test cases were generated with AI as I'm not as familiar with this test harness. Suggest verifying that it's testing properly.

### DIFF
--- a/.changeset/clever-rules-invent.md
+++ b/.changeset/clever-rules-invent.md
@@ -1,0 +1,6 @@
+---
+"@clack/prompts": minor
+"@clack/core": minor
+---
+
+Fixed spinner onCancel not being called on Ctrl+C and prevents process.exit call
--- a/packages/core/src/utils/index.ts
+++ b/packages/core/src/utils/index.ts
@@ -29,6 +29,7 @@ interface BlockOptions {
 	output?: Writable;
 	overwrite?: boolean;
 	hideCursor?: boolean;
+	onSoftCancel?: () => void;
 }
 
 export function block({
@@ -36,6 +37,7 @@ export function block({
 	output = stdout,
 	overwrite = true,
 	hideCursor = true,
+	onSoftCancel,
 }: BlockOptions = {}) {
 	const rl = readline.createInterface({
 		input,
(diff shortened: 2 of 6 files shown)


43081j:
this feels like a workaround. i'll have a deeper think about it soon but i don't think we should be introducing conditional logic here. i think if we're going to do anything, we should just have an `onCancel` that defaults to `() => process.exit(0)`.

but it needs some thought first

kajoseph:
Honestly, I would prefer the process.exit call be removed entirely as it feels kind of reckless/sloppy programming-wise, but that's functionally a breaking change for users that rely on it (though I personally was relying on the spinner returning control back to the program - did this behavior change between alpha and 1.5? I digress).

I can test `onCancel` defaulting to `() => process.exit(0)`. My sense is that anything more drastic would probably need to wait for a 2.0 release.

PR #588, opened 2026-07-25
fix(prompts): style box guide bar to match other gutters

danielroe:
## What does this PR do?

noticed this when working on https://github.com/nuxt/cli/pull/1386 - the guide for boxes was a brighter grey/white than everywhere else

it seems like a straightforward fix 🙏

I couldn't find any issue tracking

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

### DIFF
--- a/packages/prompts/test/box.test.ts
+++ b/packages/prompts/test/box.test.ts
@@ -98,6 +98,23 @@ describe.each(['true', 'false'])('box (isCI = %s)', (isCI) => {
 		expect(output.buffer).toMatchSnapshot();
 	});
 
+	test('renders the guide in the same style as log', () => {
+		const logOutput = new MockWritable();
+		prompts.log.info('message', { input, output: logOutput });
+		const [logGuide] = logOutput.buffer.join('').split('\n');
+
+		prompts.box('message', 'title', {
+			input,
+			output,
+			width: 'auto',
+			formatBorder: (text) => styleText('green', text),
+		});
+
+		for (const line of output.buffer) {
+			expect(line.startsWith(`${logGuide} `)).toBe(true);
+		}
+	});
+
 	test('renders without guide when withGuide is false', () => {
 		prompts.box('message', 'title', {
 			input,
--- a/.changeset/box-guide-colour.md
+++ b/.changeset/box-guide-colour.md
@@ -1,0 +1,5 @@
+---
+"@clack/prompts": patch
+---
+
+Style the guide bar rendered by `box({ withGuide: true })` in grey so it matches the gutter drawn by `log`, `note` and the spinner
--- a/packages/prompts/src/box.ts
+++ b/packages/prompts/src/box.ts
@@ -1,4 +1,5 @@
 import type { Writable } from 'node:stream';
+import { styleText } from 'node:util';
 import { getColumns, settings } from '@clack/core';
 import stringWidth from 'fast-string-width';
 import { wrapAnsi } from 'fast-wrap-ansi';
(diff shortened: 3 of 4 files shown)


dreyfus92 (approved):
thank you @danielroe ✌🏻

packages/prompts/test/box.test.ts
```diff
@@ -98,6 +98,23 @@ describe.each(['true', 'false'])('box (isCI = %s)', (isCI) => {
 		expect(output.buffer).toMatchSnapshot();
 	});
 
+	test('renders the guide in the same style as log', () => {
```
43081j, line 101: nit: too late now that we've merged this but i don't think this test is quite right.

we don't need to test two prompts against each other. box simply had a bug in that it wasn't rendering the guide properly. now that we've fixed that and updated the snapshots, those snapshots should catch any regression going forward.

i.e. `log` and `box` only happen to be styled the same, we haven't defined anywhere that they're forever meant to be

danielroe: isn't the idea that you can use all the utilities and have a consistent guide line? we wouldn't expect it to be a different colour based on which one you use, right?

43081j: it is. do you think `box` owns that project-wide consistency?

currently this is handled by the fact we snapshot every rendered prompt. that's how we test that they all render the right guide - and we basically missed that `box` was wrong in the initial snapshot.

now that it is right, the snapshot is what will tell us if anyone ever breaks the guide.

if we did want to double-test that, it isn't a prompt-specific test because you'll just end up intermingling prompt tests. it makes more sense, if anything, to have a generic test that goes through each prompt and checks the guide.

though i wouldn't bother since the snapshots already achieve that

43081j: something like this:

```ts
for (const p of listOfPrompts) {
  // find the first line with a "guide", add it to a set
}
// if the set has >1, test fails
```

but at that point, you're basically testing what the snapshot tests already do. though you are making it more explicit

danielroe: i was responding to this:
> i.e. log and box only happen to be styled the same, we haven't defined anywhere that they're forever meant to be

my response: isn't the idea that you can use all the utilities and have a consistent guide line? we wouldn't expect it to be a different colour based on which one you use, right?

(i'm fine if you don't want a separate test- that makes sense! - but it is unsettling that you are saying that they aren't meant to match colour wise)

43081j: do you understand the reason the test belongs elsewhere (or isn't necessary)?

on guides, yes i should have been clearer. the guide is meant to be consistent across all prompts but isn't necessarily going to be the same style. e.g. `select` styles its bar based on what part of the component is being rendered.

testing `box` against `log` is basically at the wrong level/abstraction is all. `box` isn't implementing the guide of `log`, both are implementing the standard guide. so a test to ensure they have the right guide belongs to whatever owns the "standard guide", i.e. some main file.

danielroe: I'll remember your preferences in future 👌

43081j: hmm no its not a preference 😅 its just where such a test belongs. if you disagree with that, it means i must be missing something, so please do explain.

you're using `log` as a means to test that `box` is consistent with _some other prompt_, one you hope is consistent itself. but if we're going to test consistency, we should just have a test for that which asserts that _all_ prompts are consistent. make sense? not a preference, just a standard test.

danielroe: I'm not disagreeing with you, I just don't want to take my morning debating the placement of a test.

if you're happy with just a snapshot test, that is _totally fine_.

next time I open a PR for this kind of thing in clack, I will just update the snapshots.

43081j: makes sense 👍 

to be clear though that's not the take away here. the test is fine if it was in the right place and written in a more generic way that covers all prompts.

danielroe: you said:
> though i wouldn't bother since the snapshots already achieve that.

I understand you're saying we don't need a separate test if the snapshots already cover it.

danielroe: to be clear, I'm totally happy with removing the test in this PR and don't think you're missing anything by doing so.

in a project I maintained, I would probably additionally have a logic test that asserts that colours match across prompts. I'm not familiar enough with the codebase here to write such a test but it would be better than just asserting that box matched log.

Issue #585, opened 2026-07-17
rfc: accessible mode


dreyfus92:
coming out of the discord convo about spinners and screen readers. _tldr_ of the problem: spinners and live-redrawn prompts get re-announced on every repaint by voiceover/nvda/orca, which makes clack prompts effectively unusable with assistive tech. github hit the same wall with their cli and shipped static text + numbered inputs instead: https://github.blog/engineering/user-experience/building-a-more-accessible-github-cli/

detecting screen readers isn't viable (the reader runs client-side, so over ssh/containers there's zero signal, and local platform APIs are inconsistent + amount to sniffing disability status). so the proposal is an explicit opt-in following the emerging `ACCESSIBLE` env var convention (charmbracelet/huh already honors it in go land). nothing in the js prompt ecosystem does this yet, so clack adopting it likely sets the convention.

this issue is the coordination doc: design + checklist live here, chatter can stay on discord.

## activation semantics

accessible mode is enabled when any of these resolve true, in priority order:

1. explicit `accessible: true` option on a prompt (or global setting)
2. `ACCESSIBLE` env var set to any non-empty value (`0` and `false` explicitly disable)
3. open question: should `!isTTY` imply it, or stay a separate degradation path?

## behavior per prompt

| prompt | accessible behavior |
|---|---|
| `spinner` | print message once, plain `still working: <msg>` line every 30-60s (configurable), final done/error line. zero redraws |
| `progress-bar` | log at milestones (25/50/75/100) instead of continuous repaint |
| `task` / `task-log` / `stream` | audit for line rewrites, route through same static path |
| `select` / `select-key` | numbered list, read a number from stdin, re-prompt on invalid |
| `multi-select` / `group-multi-select` | numbered list, comma-separated numbers in, echo back parsed selection |
| `limit-options` | no pagination, print full list |
| `autocomplete` | degrade to plain "type your answer" + printed options (may split to follow-up) |
| `text` / `confirm` / `password` / `path` / `multi-line` / `date` | strip cursor-repositioning decorations, plain question in / plain input out, confirm becomes `(y/n)` |
| `note` / `box` / `log` / `messages` | plain text with clear begin/end delimiters, no box-drawing repaints |

## implementation approach

intent is for accessible mode to be a distinct render path rather than `if (accessible)` conditionals scattered through every prompt. all prompts already flow through the shared render layer in core, so the accessible output should hang off that seam. keeping it isolated there also means future renderer work inherits it for free.

## proposed plan

[...]

PR #595, opened 2026-08-13
feat(core): add `accessible` prompt option

dreyfus92:
## What does this PR do?

this threads the `accessible` option through the shared prompt options so every prompt can opt into accessible mode individually. 

tests cover the resolution order from the prompt's side.

related: #585

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -7,17 +7,6 @@ import type { Action } from './utils/settings.js';
 export type ClackState = 'initial' | 'active' | 'cancel' | 'submit' | 'error';
 
 /**
- * Options shared by all Clack prompts
- */
-export interface CommonPromptOptions {
-	/**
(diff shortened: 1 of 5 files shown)


packages/core/src/types.ts
```diff
@@ -6,6 +6,17 @@ import type { Action } from './utils/settings.js';
  */
 export type ClackState = 'initial' | 'active' | 'cancel' | 'submit' | 'error';
 
+/**
+ * Options shared by all Clack prompts
```
43081j: isn't this just `PromptOptions`? maybe we don't need to extract it into its own interface. `PromptOptions` is already a base

dreyfus92: you're right, nice catch james 😭

43081j (approved):

PR #596, opened 2026-08-20
feat(prompts): add accessible mode to `spinner`

dreyfus92:
## What does this PR do?

adds accessible mode to `spinner` (pr 1 of #585).

when enabled (`accessible` option > global setting > `ACCESSIBLE` env var), the spinner emits static append-only output instead of animated repaints:

- start message printed once, plain text
- `still working: <msg>` heartbeat every 30s (`accessibleInterval`, 0 disables)
- plain final line (`Done` / cancel / error), no glyphs, no ansi

baseline voiceover recordings showed repaints are completely silent to VO and glyphs get read out as words. cancellation/signal semantics unchanged, accessible takes precedence over CI mode.

Closes #

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

### DIFF
--- a/packages/prompts/src/spinner.ts
+++ b/packages/prompts/src/spinner.ts
@@ -1,5 +1,5 @@
 import { styleText } from 'node:util';
-import { block, getColumns, settings } from '@clack/core';
+import { block, getColumns, isAccessible, settings } from '@clack/core';
 import { wrapAnsi } from 'fast-wrap-ansi';
 import { cursor, erase } from 'sisteransi';
 import {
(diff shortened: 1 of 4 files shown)


packages/prompts/src/spinner.ts
```diff
@@ -20,6 +20,12 @@ export interface SpinnerOptions extends CommonOptions {
 	frames?: string[];
 	delay?: number;
 	styleFrame?: (frame: string) => string;
+	/**
+	 * Milliseconds between "still working" heartbeat lines in accessible mode.
+	 * Set to `0` to disable the heartbeat.
+	 * @default 30_000
+	 */
+	accessibleInterval?: number;
```
43081j, line 28: who is expected to configure this? does it need to be configurable?

we should avoid adding noise to the public API if we don't need to

dreyfus92: it came from one of the open questions in the rfc issue, the interval default for the status lines, 30s or 60s. no one picked one so i made it configurable and defaulted to 30s. 

i can fix this to be set to 30s.

umfarah1406: 111iixza

43081j (approved):

PR #599, opened 2026-09-06
fix: output submit gap only when guide enabled

### LINKED ISSUE #598: [Bug] text: submitted value line is indented 2 spaces while active input is flush-left under withGuide: false (content shifts right on submit)
## Environment

- `@clack/prompts` 1.7.0 (also reproduced against current `main` rendering logic)
- Node 22, macOS, `TERM=xterm-256color`, 80-col PTY + xterm-headless grid
- CJK input used in the example, but the shift happens with ASCII input too (any text prompt value)

## Minimal repro

```js
import * as clack from '@clack/prompts'

clack.updateSettings({ withGuide: false })
await clack.text({ message: '插件包名' })  // type 测, then press Enter
```

## Observed

While the prompt is **active**, the input row is rendered flush-left:

```
◆  插件包名
测█        ← cursor block right after the text, column 0 + displayWidth
```

After pressing **Enter**, the submitted (final) frame re-renders the value line with a 2-space indent:

```
◇  插件包名
  测       ← shifted right by 2 columns
```

So with `withGuide: false`, every text submit makes the content visibly jump 2 columns to the right.

With the default `withGuide: true`, both renderings carry the same `│  ` prefix (3 columns), so the frames are consistent — the inconsistency only appears when the guide is turned off, because the submit branch keeps a 2-space indent (the leftover of the old `│  ` prefix) while `userInputWithCursor` renders flush-left.

## Byte streams (PTY capture, 80 cols)

Last redraw while typing `测` (active state, flush-left):

```
\x1b[999D\x1b[3A\x1b[1B\x1b[2K\x1b[G测█\x1b[2B
```

Full output after Enter (submit frame, 2-space-indented value line):

```
\x1b[999D\x1b[3A\x1b[J\x1b[32m◇\x1b[39m  插件包名\r\n  \x1b[2m测\x1b[22m\r\n\x1b[?25h
```

The `  ` before the dimmed value in the submit frame is the discrepancy: no such prefix exists in the active render.

## Expected

Under `withGuide: false`, the submitted value line should be rendered flush-left like the active input row (i.e. no 2-space carry-over), so the frame doesn't shift on submit.

## Notes

Found while building a PTY-based CJK cursor regression harness; happy to send a PR if the fix direction (skip the 2-space indent on the submit branch when `settings.withGuide === false`) sounds right.

43081j:
In the text prompt, we output a `"  "` prefix to align the submitted
value with the guide. When the guide is disabled, this was leaving is
with an unnecessary gap.

This fix removes that gap.

Fixes #598.

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

### DIFF
--- a/packages/core/test/prompts/prompt.test.ts
+++ b/packages/core/test/prompts/prompt.test.ts
@@ -467,6 +467,7 @@ describe('Prompt', () => {
 		expect(eventSpy).not.toHaveBeenCalled();
 		expect(instance.state).to.equal('validating');
 
+		// biome-ignore lint/style/noNonNullAssertion: We know resolveValidation is defined
 		resolveValidation!(undefined);
 		await resultPromise;
 
--- a/.changeset/hungry-rice-film.md
+++ b/.changeset/hungry-rice-film.md
@@ -1,0 +1,5 @@
+---
+"@clack/prompts": patch
+---
+
+Fixes a withGuide rendering issue leaving extra spaces in text prompts.
--- a/packages/prompts/src/text.ts
+++ b/packages/prompts/src/text.ts
@@ -91,7 +91,7 @@ export const text = (opts: TextOptions) => {
 					return `${title.trim()}\n${errorPrefix}${userInput}\n${errorPrefixEnd}${errorText}\n`;
 				}
 				case 'submit': {
-					const valueText = value ? `  ${styleText('dim', value)}` : '';
+					const valueText = value ? `${hasGuide ? '  ' : ''}${styleText('dim', value)}` : '';
 					const submitPrefix = hasGuide ? styleText('gray', S_BAR) : '';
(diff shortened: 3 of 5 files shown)


packages/core/test/prompts/prompt.test.ts
```diff
@@ -467,7 +467,7 @@ describe('Prompt', () => {
 		expect(eventSpy).not.toHaveBeenCalled();
 		expect(instance.state).to.equal('validating');
 
-		resolveValidation!(undefined);
+		resolveValidation?.(undefined);
```
43081j: biome kept changing this to an optional chain but then tsc complains since it hasn't been initialised. so i just ignored it

dreyfus92 (approved):
sweet! thanks @43081j 🫡

PR #601, opened 2026-09-08
fix: correctly specify return type of `CANCEL_SYMBOL`

### LINKED ISSUE #600: [Bug] isCancel no longer narrows prompt results after 1.5.0
**Environment**

- OS: macOS
- Node Version: v24.20.0
- Package: `@clack/core`, `@clack/prompts`
- Package Version: core 1.5.0, prompts 1.8.0

**Describe the bug**

#592 changed `isCancel` from `value is symbol` to `value is typeof CANCEL_SYMBOL`, but every prompt still returns `Promise<T | symbol>` (core `Prompt.prompt()` and the 11 `as Promise<... | symbol>` casts in prompts). typescript can't subtract a unique symbol from `symbol`, so after the guard the result still has `symbol` in its type and any downstream use errors. this breaks code that type-checked fine on 1.7.0.

**To Reproduce**

```ts
import { path, isCancel } from '@clack/prompts';

const p = await path({ message: 'pick a file' });
if (isCancel(p)) process.exit(0);
console.log(`selected: ${p}`);
// TS2731: Implicit conversion of a 'symbol' to a 'string' will fail at runtime.
```

same snippet passes on `prompts 1.7.0` / `core 1.4.3`. assigning the result to a string fails with TS2322 instead.

**Expected behavior**

`isCancel` narrows the cancel case away like it did before. #554 asked for the specific cancel type on prompt return types, and #592 only did the guard half, which is what left the two out of sync.

gameroman:
## What does this PR do?

Closes #600

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Documentation
- [ ] Performance improvement
- [ ] Tests

## Checklist

- [ ] `pnpm test` passes (or targeted tests for my change)
- [ ] `pnpm format` has been run
- [ ] I have added a changeset

### DIFF
--- a/.changeset/lucky-plums-tell.md
+++ b/.changeset/lucky-plums-tell.md
@@ -1,0 +1,6 @@
+---
+"@clack/core": patch
+"@clack/prompts": patch
+---
+
+Return type for prompts now correctly specifies `CANCEL_SYMBOL` instead of `symbol`
--- a/packages/core/src/prompts/prompt.ts
+++ b/packages/core/src/prompts/prompt.ts
@@ -144,7 +144,7 @@ export default class Prompt<TValue> {
 	}
 
 	public prompt() {
-		return new Promise<TValue | symbol | undefined>((resolve) => {
+		return new Promise<TValue | typeof CANCEL_SYMBOL | undefined>((resolve) => {
 			if (this._abortSignal) {
 				if (this._abortSignal.aborted) {
 					this.state = 'cancel';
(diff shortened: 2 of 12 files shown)


dreyfus92 (approved):
this looks good! but, could we add a type-level regression test? something like
```ts
if (isCancel(value)) return;
expectTypeOf(value).toEqualTypeOf<string>();
```
inside an existing text or path test. tsc passed on the 1.8.0 base, so nothing today would fail if the narrowing drifts again. expectTypeOf is already available from vitest and the test dirs are typechecked in CI.

cc @43081j

43081j (approved):

PR #608, opened 2026-09-18
fix: treat IME-committed ideographic space as the space key

kai-natori:
## What does this PR do?

Japanese (and other CJK) IMEs commit an ideographic space (`U+3000`) instead of `U+0020` when the space key is pressed. Node readline reports such keypresses with no `key.name`, so prompts never see a space key — actions like toggling multiselect options silently do nothing while an IME is active.

This PR normalizes a keypress whose `key.name` is undefined and whose committed char is `U+3000` to `name: 'space'` in the base `Prompt.onKeypress`, fixing the behavior for multiselect, group multiselect, autocomplete, and custom prompts. Text input is unaffected — the committed character still lands in the readline buffer as before.

No linked issue — I ran into this directly while using clack-based prompts with Japanese IME input. Happy to file one if preferred.

## Type of change

- [x] Bug fix

## Checklist


packages/core/src/prompts/prompt.ts
```diff
@@ -229,6 +229,13 @@ export default class Prompt<TValue> {
 			return;
 		}
 
+		// Some IMEs (e.g. Japanese input) commit an ideographic space (U+3000)
+		// instead of U+0020 when the space key is pressed. Treat it as the space
+		// key so key-driven actions like toggling options still work.
+		if (key && key.name === undefined && char === '　') {
```
43081j: what if you want to catch U+3000?

unfortunately i think this has to live in the consuming code rather than the core. i.e. update multi-select to explicitly look for this

kai-natori: Good point — moved the check into MultiSelectPrompt and GroupMultiSelectPrompt so U+3000 stays observable for consumers.

packages/core/test/prompts/group-multiselect.test.ts
```diff
@@ -30,6 +30,24 @@ describe('GroupMultiSelectPrompt', () => {
...
+			options: {
+				group: [{ value: 'foo' }, { value: 'bar' }],
+			},
+			selectableGroups: false,
+		});
+		instance.prompt();
+
+		input.emit('keypress', '　', { sequence: '　' });
+		expect(instance.value).toEqual(['foo']);
+		input.emit('keypress', '　', { sequence: '　' });
```
43081j:
```suggestion
		input.emit('keypress', '　', { sequence: '\u3000' });
		expect(instance.value).toEqual(['foo']);
		input.emit('keypress', '　', { sequence: '\u3000' });
```

packages/core/test/prompts/multi-select.test.ts
```diff
@@ -143,6 +143,21 @@ describe('MultiSelectPrompt', () => {
...
+				input,
+				output,
+				render: () => 'foo',
+				options: [{ value: 'foo' }, { value: 'bar' }],
+			});
+			instance.prompt();
+
+			input.emit('keypress', '　', { sequence: '　' });
+			expect(instance.value).toEqual(['foo']);
+			input.emit('keypress', '　', { sequence: '　' });
```
43081j:
```suggestion
			input.emit('keypress', '　', { sequence: '\u3000' });
			expect(instance.value).toEqual(['foo']);
			input.emit('keypress', '　', { sequence: '\u3000' });
```

packages/core/src/prompts/group-multiselect.ts
```diff
@@ -67,6 +67,15 @@ export default class GroupMultiSelectPrompt<T extends { value: any }> extends Pr
 			this.#selectableGroups ? 0 : 1
 		);
 
+		this.on('key', (char, key) => {
+			// Some IMEs (e.g. Japanese or Chinese input) commit an ideographic space (U+3000)
+			// with no key name when the space key is pressed. Treat it as space so
+			// toggling options still works.
```
43081j:
```suggestion
			// Treat ideographic spaces as if they were normal spaces.
```

43081j (changes requested):

43081j (approved):

